PR #6, opened 2017-07-04
Read software version from `version`

aschmolck:
Easier to bump etc.


setup.py
```diff
@@ -1,7 +1,7 @@
 from distutils.core import setup
 setup(
     name='marge',
-    version='0.1',
+    version=open('version').read().strip(),
     license='BSD3',
```
jcpetruzza: @aschmolck It seems the `open('version')` doesn't work. CWD may not be marge's directory. Something like `open(os.path.dirname(__file__) + 'version').read().strip()` may work.

PR #10, opened 2017-07-08
Various fixes

aschmolck:
Not lying about the "various". This should really be split up into ~4 different PRs, but at least the individual commits should be fairly well organized.

### DIFF
--- a/dockerize.nix
+++ b/dockerize.nix
@@ -1,4 +1,4 @@
-{ pkgs ? import <nixpkgs> {} }:
+{ pkgs ? import ./pinnedNixpkgs.nix }:
 with pkgs;
 let callPackage = pkgs.lib.callPackageWith (pkgs);
   marge = callPackage ./marge.nix {};
--- a/marge/gitlab.py
+++ b/marge/gitlab.py
@@ -1,4 +1,5 @@
 import json
+import logging as log
 from collections import namedtuple
 
 import requests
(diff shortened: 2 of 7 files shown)


dockerize.nix
```diff
@@ -11,7 +11,8 @@ dockerTools.buildImage {
   ${dockerTools.shadowSetup}
   mkdir -p /root/.ssh
   '';
-  contents = [marge pkgs.bash pkgs.coreutils pkgs.openssh];
+  # gnu sed and grep are needed for git filter-branch; in other words pkgs.git seems incomplete
+  contents = [marge pkgs.bash pkgs.coreutils pkgs.openssh pkgs.gnused pkgs.gnugrep];
   config = {
```
jcpetruzza: @aschmolck Adding `sed` and `grep` here is pointless, right? If so, d8299af86462b7efe could be dropped.

marge/gitlab.py
```diff
@@ -1,21 +1,26 @@
...
 
 import requests
 
 
 class Api(object):
-    def __init__(self, gitlab_url, auth_token):
+    def __init__(self, gitlab_url, auth_token, debug=False):
         self._auth_token = auth_token
         self._api_base_url = gitlab_url.rstrip('/') + '/api/v4'
+        self._debug = debug
```
jcpetruzza: @aschmolck We don't seem to be using `self.debug` anywhere, I'd just drop it

Issue #19, opened 2017-07-18
TypeError: 'NoneType' object is not subscriptable


NLincoln:
We're trying to get marge-bot up and running on our own gitlab instance, and might have found a bug?
```
> marge/project.py
>> return AccessLevel(self.info['permissions']['project_access']['access_level'])
>> TypeError: 'NoneType' object is not subscriptable
```

After looking into this a bit, it looks like gitlabs api will return permissions only on the `/projects/:id` request, but not for `/projects`.

jcpetruzza:
Thanks for reporting this! What version of GitLab are you on? Also, if you launch marge-bot with `--debug` you should get the GitLab's REST API responses, could you paste that as well?

jcpetruzza:
Looking at the docs the [docs](https://docs.gitlab.com/ce/api/projects.html#list-projects), their example call to `GET /projects` returns a list with two projects, one includes the `permissions` field, the other doesn't. Trying to figure out the logic...

NLincoln:
It looks like only the call that gets a single project will return permissions.

jcpetruzza:
The behaviour of the GitLab REST API can change with versions and this can be the reason why we see different results. What version of GitLab are you? You should be able to see it under `Help` or `Settings`

NLincoln:
9.3.5 ee

jcpetruzza:
Ok, that's a recent one.

What are the permissions the `@marge-bot` user has on your project? (e.g. `Guest`, `Reporter`, `Developer`, `Master`?).

Also, when you run it with `--debug` and get the list of projects for `GET /projects`, do you see many fields per project or just a few? (the REST call accepts a `simple` arg that may be defaulting to `True` in your version of GitLab).

NLincoln:
This is curious... after looking at the debug output gitlab _is_ sending the `permissions` field.

NLincoln:
Sometimes though the `permissions.project_access` can be null. I think that's the issue here

NLincoln:
We use group level permissions for our users (since we have several projects). Could that be part of it?

david-duncan:
Believe it or not but some projects the user has access to come back with an empty permissions array:
```
    'permissions': {
      'project_access': None,
      'group_access': None
    }
```
Even though the user is a master on the group. We tracked this down to a gitlab bug that we have not reported(or found the issue tracking) where subgroups getting moved do not cascade proper permissions down.

So we removed our user from these groups and manually re-added. Now we still have group level access
[...]

PR #20, opened 2017-07-18
#19 - NoneType on project level access. Fall back to group level acce…

david-duncan:
…ss. Handle none access more gracefully


jcpetruzza:
Great, thanks for looking into this!

PR #22, opened 2017-07-19
Add k8s templates

david-duncan:
This was from the discussion here:
https://github.com/smarkets/marge-bot/issues/19#issuecomment-316229780

There are many different ways of doing this in k8s... Helm may be better. I tend to like ktmpl in our org because we can just quickly template without a lot of boiler.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,18 @@ docker run
 
 See below for the meaning of the additional flags.
 
+Kubernetes templating with ktmpl:
+```bash
+ktmpl ./deploy.yml \
+--parameter APP_NAME "marge-bot" \
+--parameter APP_IMAGE "smarketshq/marge-bot" \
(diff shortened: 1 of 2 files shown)


aschmolck:
We've been using helm at smarkets, but I'll give this a try as well – certainly adding either a helm or ktempl template coming with the project would be quite nice. Thanks for submitting this!

aschmolck:
@david-duncan sorry for the delay, only got around to testing it yesterday. It would be nice to parameterize the optional flags, too, but ktmpl unfortunately doesn't support lists, or list interpolation, so there seems to be no elegant way of doing it. I'm going to rebase this in as is and maybe tweak it later. Thanks for your contribution!

david-duncan:
No problems. If you decide to delete it and go ahead with helm let me know and I'd be glad to contribute

PR #23, opened 2017-07-19
Prepare for 0.1.1 release

jcpetruzza:
- Update version file and changelog
  - Remove references to 0.1.0 from README
  - Minor code clean up to isolate GitLab issues from the rest of the code


david-duncan:
Thanks for cleaning up my contribution :+1:

README.md
```diff
@@ -90,7 +90,7 @@ image we provide:
 docker run
   -e MARGE_AUTH_TOKEN="$(cat marge-bot.token)" \
   -e MARGE_SSH_KEY="$(cat marge-bot-ssh-key)" \
-  smarketshq/marge-bot:0.1.0 \
+  smarketshq/marge-bot \
```
aschmolck: Don't think this has the right semantics, unless you bump latest as well (just creating an new tag w/ make dockerize won't, I think).

jcpetruzza: That makes sense when you are running the image you built yourself. I expect most people will want to use the latest image on docker hub and we should bump `latest` there as well.

aschmolck: Yup, agreed. Just saying that a) we should add something to the Makefile or whatever that does it automatically, and b) until we do, you will need to tag manually with latest (as well) before pushing.

jcpetruzza: Agreed, I think I misread your comment

PR #27, opened 2017-07-23
Add `--project-regexp` flag

aschmolck:
Main use case is running multiple instances of marge with different flags for different projects on the same repo. Beyond that it's mostly handy for debugging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,20 @@ prevent shipping late on a Friday, but still want to allow marking merge request
 More than one embargo period can be specified. Any merge request assigned to her
 during an embargo period, will be merged in only once all embargoes are over.
 
+## Restricting the list of projects marge-bot considers
+
+By default marge-bot will work on all projects that she is a member of.
+Sometimes it is useful to restrict a specific instance of marge-bot to a subset
+of projects. You can specify a regexp that projects must match (anchored at the
+start of the string) with `--project-regexp`.
+
+One use-case is if you want to use different configurations (e.g.
+--add-reviewers on one project, but not the others). A simple way of doing is
+run two instances of marge-bot passing `--add-reviewers --project-regexp
+project/with_reviewers` to the first instance and `--project-regexp
+(?!project/with_reviewers)` to the second ones. The latter regexp is a negative
+look-ahead and will match any string not starting with `project/with_reviewers`.
+
 ## Troubleshooting
 
 Marge-bot continuously logs what she is doing, so this is a good place to look
--- a/marge/app.py
+++ b/marge/app.py
@@ -6,6 +6,7 @@ import argparse
 import contextlib
 import logging
 import os
+import re
 import sys
 import tempfile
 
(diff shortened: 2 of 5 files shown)


marge/app.py
```diff
@@ -63,9 +64,20 @@ def _parse_args(args):
...
     arg('--debug', action='store_true', help='Debug logging (includes all HTTP requests etc.)')
 
-    return parser.parse_args(args)
+    options = parser.parse_args(args)
+    try:
+        options.project_regexp = re.compile(options.project_regexp)
+    except re.error as err:
+        print('ERROR: Invalid regexp: %r (%s)' % options.regexp, err.msg, file=sys.stderr)
+        sys.exit(1)
+    return options
```
jcpetruzza: @aschmolck Why not do something like this?
```python
def regexp(re_str):
    try:
        return re.compile(re_str)
    except re.error as err:
        raise argparse.ArgumentTypeError('Invalid regexp: {re_str!r} ({err.msg})'.format(re_str=re_str, err=err))

# .... and then
    arg(
        '--re-project-regexp',
        type=regexp,
        default='.*',
        help=...,
    )
```

marge/bot.py
```diff
@@ -61,8 +64,18 @@ def _run(self, repo_manager):
...
+            filtered_out = set(my_projects) - set(filtered_projects)
+            if filtered_out:
+                log.info(
+                    'Projects that match project_regexp: %s',
+                    [p.path_with_namespace for p in filtered_projects]
+                )
+                log.info(
+                    'Projects that do not match project_regexp: %s',
+                    [p.path_with_namespace for p in filtered_out]
+                )
```
jcpetruzza: @aschmolck This is valuable info to have, but I would only print it under `--debug`

README.md
```diff
@@ -177,6 +177,20 @@ prevent shipping late on a Friday, but still want to allow marking merge request
 More than one embargo period can be specified. Any merge request assigned to her
 during an embargo period, will be merged in only once all embargoes are over.
 
+## Restricting the list of projects marge-bot considers
+
+By default marge-bot will work on all projects that she is a member of.
+Sometimes it is useful to restrict a specific instance of marge-bot to a subset
+of projects. You can specify a regexp that projects must match (anchored at the
+start of the string) with `--project-regexp`.
+
```
jcpetruzza: @aschmolck I understand the "anchored at the start of the string" part as saying that the there  is an implicit `^` at the beginning of the regexp, so that `--project-regexp foo/bar` will not match project `cafoo/bar`. Is that the intended meaning? If so, where does this happen in the code?

marge/bot.py
aschmolck: No printed by default (orginal version was log.debug, but I've modified it to log.info on the basis that log.debug is too noisy in the general case, and this should not be very noisy unless run on gitlab.com or somesuch).

README.md
aschmolck: Yup, your understanding is correct and it's the `self.project_regexp.match` that does it (the alternative would be `self.project_regexp.search`, [which does not anchor at the beginning](https://docs.python.org/2/library/re.html#re.RegexObject)).

marge/app.py
aschmolck: Just laziness.

README.md
jcpetruzza: :+1:

marge/bot.py
jcpetruzza: Sorry, I don't follow here. This is currently going to log the filtered out projects continuously (it is under a `while True`) and because it is very unlikely that a project is renamed, all except the first occurrence will be redundant. I would just put everything from `filtered_out = ...` under an `if self._debug:` or something like that.

Issue #30, opened 2017-07-25
Missing support for gitlab-CE


backfx:
Looks like marge-bot is only compatible with gitlab-EE. It uses the approval feature that is not available in gitlab-CE. The endpoint `/api/v4/projects/{id}/merge_requests/{id}/approvals` is also not part of the Community Editions API.

```25.7.2017 10:23:142017-07-25 08:23:14,388 DEBUG Starting new HTTP connection (1): gitlab
25.7.2017 10:23:142017-07-25 08:23:14,400 DEBUG http://gitlab:80 "GET /api/v4/projects/1/merge_requests/1/approvals HTTP/1.1" 404 None
25.7.2017 10:23:142017-07-25 08:23:14,401 DEBUG RESPONSE CODE: 404
25.7.2017 10:23:142017-07-25 08:23:14,401 DEBUG RESPONSE BODY: {'error': '404 Not Found'}
25.7.2017 10:23:142017-07-25 08:23:14,402 ERROR Unexpected Exception
25.7.2017 10:23:14Traceback (most recent call last):
25.7.2017 10:23:14  File "/nix/store/02cbi06m7hk3qccalxrc71c2abiax6nz-python3.6-marge-0.1.2/lib/python3.6/site-packages/marge/job.py", line 56, in execute
25.7.2017 10:23:14    approvals = merge_request.fetch_approvals()
25.7.2017 10:23:14  File "/nix/store/02cbi06m7hk3qccalxrc71c2abiax6nz-python3.6-marge-0.1.2/lib/python3.6/site-packages/marge/merge_request.py", line 121, in fetch_approvals
25.7.2017 10:23:14    approvals.refetch_info()
25.7.2017 10:23:14  File "/nix/store/02cbi06m7hk3qccalxrc71c2abiax6nz-python3.6-marge-0.1.2/lib/python3.6/site-packages/marge/approvals.py", line 15, in refetch_info
25.7.2017 10:23:14    self._info = self._api.call(GET(approver_url))
25.7.2017 10:23:14  File "/nix/store/02cbi06m7hk3qccalxrc71c2abiax6nz-python3.6-marge-0.1.2/lib/python3.6/site-packages/marge/gitlab.py", line 55, in call
25.7.2017 10:23:14    raise error(response.status_code, err_message)
25.7.2017 10:23:14marge.gitlab.NotFound: (404, {'error': '404 Not Found'})
```

Supporting both versions with suitable workflows would be awesome!

[...]

PR #31, opened 2017-07-26
Prevent calls to approvals endpoint on CE

jcpetruzza:
The `/projects/id/merge_requests/iid/approvals` endpoint is only available in the Enterprise Edition. On CE, we can instead pretend that there are no approvals left.

See #30.


aschmolck:
LGTM.

PR #36, opened 2017-08-07
readme: add snippet to create key into env var

glensc:
did not know good place to put this, or how to word it. just wanted to get it out there.


glensc:
disregard this. docker-compose .env file does not support any parser

PR #40, opened 2017-08-07
Better handling of temporary ssh-key files

### LINKED ISSUE #38: stale ssh private key files
the current version (59ace57) leaves temporary ssh key files laying around. 

first problem is that the temporary file is [created](https://github.com/smarkets/marge-bot/blob/59ace5743635307046ae4053418fea59bd470015/marge/app.py#L94) even if `--ssh-key-file` commandline is used.

and second problem that the temporary files are never removed. thus root filesystem of docker container is filled with ssh private key files (probably system temp dir for non-docker runs).

```
bash-4.4# ls -l /
-rw-------   1 root root 3243 Aug  7 09:07 ssh-key
-rw-------   1 root root    0 Aug  7 10:36 ssh-key-rgss44m3
-rw-------   1 root root    0 Aug  7 10:35 ssh-key-tmagyr7i
```

`ssh-key` is mounted via docker volume and used as commandline `--ssh-key-file=/ssh-key`, the other two came from processing (two?) merge requests.

jcpetruzza:
This solves two issues:

  1. Unnecessary ssh-key-xxxx files are created when a proper key file was given
  2. The former files are never removed.

Notice that the second happens since `NamedTemporaryFile` needs the file
to be explicitly closed before deleting it.

Fixes #38

### DIFF
--- a/marge/app.py
+++ b/marge/app.py
@@ -96,12 +96,14 @@ def _secret_auth_token_and_ssh_key(options):
         yield auth_token.strip(), ssh_key_file
     else:
         with tempfile.NamedTemporaryFile(mode='w', prefix='ssh-key-') as env_ssh_key_file:
-            ssh_key = os.getenv('MARGE_SSH_KEY')
-            assert ssh_key, "You need to pass --ssh-key-file or set envvar MARGE_SSH_KEY"
-            env_ssh_key_file.write(ssh_key + '\n')
-            env_ssh_key_file.flush()
-            yield auth_token.strip(), env_ssh_key_file.name
-            env_ssh_key_file.close()
+            try:
+                ssh_key = os.getenv('MARGE_SSH_KEY')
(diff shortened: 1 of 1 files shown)


marge/app.py
```diff
@@ -91,15 +91,19 @@ def _secret_auth_token_and_ssh_key(options):
...
-    with tempfile.NamedTemporaryFile(mode='w', prefix='ssh-key-') as env_ssh_key_file:
-        ssh_key_file = options.ssh_key_file
-        if not ssh_key_file:
-            ssh_key = os.getenv('MARGE_SSH_KEY')
-            assert ssh_key, "You need to pass --ssh-key-file or set envvar MARGE_SSH_KEY"
-            env_ssh_key_file.write(ssh_key + '\n')
-            env_ssh_key_file.flush()
-            ssh_key_file = env_ssh_key_file.name
+    ssh_key_file = options.ssh_key_file
+    if ssh_key_file:
```
aschmolck: I don't like the use of the temporary file being conditional. Why not always create a temporary file?

PR #44, opened 2017-08-13
Make CI waiting time configurable (+ option refactoring)

jcpetruzza:
At the moment, it was hard-coded that marge would wait 15 minutes for CI to finish. Different teams have different needs, so now this parameter can be set with  `--max-ci-time-in-minutes`, which defaults to 15.

In doing so, this PR also refactors the handling of configuration parameters, so that we don't need to keep an ever-growing list of arguments in the Bot constructor. In the process, we remove the dependency of `MergeJob` on `Bot` and modify the `--embargo` parameter so that it can now receive all the embargo times separated by commas.

### DIFF
--- a/marge/app.py
+++ b/marge/app.py
@@ -63,11 +63,9 @@ def _parse_args(args):
     )
     arg(
         '--embargo',
-        type=str,
-        action='append',
-        metavar='INTERVAL',
-        default=[],
-        help='Time during which no merging is to take place, e.g. "Friday 1pm - Monday 9am".',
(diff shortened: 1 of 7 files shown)


marge/app.py
```diff
@@ -136,9 +136,7 @@ def main(args=sys.argv[1:]):
             add_tested=options.add_tested,
```
glensc: `s/now/no/` in commit message?

jcpetruzza: "No", "now" and "know" in the same sentence, it was bound to happen...

Issue #14, opened 2017-07-14
Handle autosquash


aschmolck:
Ticking autosquash seems to break marge at moment.

jcpetruzza:
I've investigated the problematic merge-request that lead to this issue being opened and it turns out auto-squash was a red-herring: the actual problem was actually another instance of #32.

It was good anyway that it made us notice we needed something like #45 to be implemented.

aschmolck:
@jcpetruzza yeah, I know should have updated the bug description. It's still  kinda true though, it interacts badly with any of the `--add*` flags.

PR #45, opened 2017-08-13
Refuse to merge if auto-squash and tagged commits

jcpetruzza:
This is a very simple of dealing with #14: just refuse to merge stuff if auto-squash is enabled and we want to tag the final commit.

In the future, a better approach could be to edit the merge request  description and put the tags there (assuming that this then becomes the full commit message, would need to try the feature more).


glensc:
what do you mean by tag? git tag? can marge create git tag after merge? did not find that from readme.

jcpetruzza:
We meant tagging a commit with `Reviewed-by:`, `Tested-by:`, etc.

PR #46, opened 2017-08-13
Better handling of failures when trying to merge

### LINKED ISSUE #32: Handle GitLab setting merge request as WIP after pushing
Because of [this GitLab feature](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/8124), we can have the following interaction:

  1. Someone creates a merge request containing a commit with WIP in its name. GitLab doesn't consider the Merge Request as WIP at this point.
  1. Merge request is assigned to Marge; it checks that the merge requests is not WIP and proceeds.
  1 The branch is rebased and pushed again.
  1. Because we have just pushed a commit that contains WIP (even if it was in one of the original commits), GitLab marks the merge request as WIP (it even reports "Marge Bot marked as Work In Progress").
  1. After CI passes, she tries to merge, but GitLab now refuses to merge and we fail with "had some issue with gitlab"

We should test for WIP status before merging or, perhaps better, after the merge failed and we don't know why.

### LINKED ISSUE #33: Enabling automerge on gitlab side during marge processing causes exceptional behavior
Repro:

1) Assign marge
2) During marges processing enable automerge
3) Sometime later marge will error out: "I'm broken on the inside, please somebody fix me."

Everything worked but someone was impatient so marge complains

jcpetruzza:
These commits improve the messages that marge-bot will leave on a merge-request when a merge attempts succeeds, based on cases we've seen in the past. E.g. this fixes #32 and fixes #33.

### DIFF
--- a/marge/job.py
+++ b/marge/job.py
@@ -178,6 +178,13 @@ class MergeJob(object):
                     )
                 elif merge_request.state == 'closed':
                     raise CannotMerge('Someone closed the merge request while I was attempting to merge it.')
+                elif merge_request.state == 'merged':
+                    # We are not covering any observed behaviour here, but if at this
+                    # point the request is merged, our job is done, so no need to complain
+                    log.info('Merge request is already merged, someone was faster!')
+                    rebased_into_up_to_date_target_branch = True
+                else:
+                    raise CannotMerge("Gitlab refused to merge this request and I don't know why!")
             except gitlab.ApiError:
                 log.exception('Unanticipated ApiError from Gitlab on merge attempt')
                 raise CannotMerge('had some issue with gitlab, check my logs...')
--- a/marge/job.py
+++ b/marge/job.py
@@ -176,6 +176,8 @@ class MergeJob(object):
                         'GitLab refused to merge this branch. I suspect that a Push Rule or a git-hook '
                         'is rejecting my commits; maybe my email needs to be white-listed?'
                     )
+                elif merge_request.state == 'closed':
+                    raise CannotMerge('Someone closed the merge request while I was attempting to merge it.')
             except gitlab.ApiError:
                 log.exception('Unanticipated ApiError from Gitlab on merge attempt')
                 raise CannotMerge('had some issue with gitlab, check my logs...')
--- a/marge/job.py
+++ b/marge/job.py
@@ -164,6 +164,13 @@ class MergeJob(object):
                 else:
                     log.warning('For the record, merge request state is %r', merge_request.state)
                     raise
+            except gitlab.MethodNotAllowed as e:
+                log.warning('Not Allowed!: %s', e)
+                merge_request.refetch_info()
+                if merge_request.work_in_progress:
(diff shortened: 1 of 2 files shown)


marge/job.py
```diff
@@ -151,6 +151,37 @@ def rebase_and_accept(self, approvals):
             except gitlab.Unauthorized:
                 log.warning('Unauthorized!')
                 raise CannotMerge('My user cannot accept merge requests!')
+            except gitlab.NotFound as e:
+                log.warning('Not Found!: %s', e)
+                merge_request.refetch_info()
+                if merge_request.state == 'merged':
+                    # someone must have hit "merge when build succeeds" and we lost the race,
+                    # the branch is gone and we got a 404. Anyway, our job here is done.
```
aschmolck: Maybe log it when it happens? Marge got blamed for missing Reviewer-tags before because she was assigned but someone pressed merge :)

(Maybe worth moving the log line above down into the `then` arm and be more specific and adapt else branch, too)

marge/job.py
```diff
@@ -151,6 +151,37 @@ def rebase_and_accept(self, approvals):
...
+                    )
+                elif merge_request.state == 'reopened':
+                    raise CannotMerge(
+                        'GitLab refused to merge this branch. I suspect that a Push Rule or a git-hook '
+                        'is rejecting my commits; maybe my email needs to be white-listed?'
+                    )
+                elif merge_request.state == 'closed':
+                    raise CannotMerge('Someone closed the merge request while I was attempting to merge it.')
+                elif merge_request.state == 'merged':
+                    # We are not covering any observed behaviour here, but if at this
```
aschmolck: again, maybe add a log line?

marge/job.py
jcpetruzza: By logging you mean "leaving a message on the MR"? If we are executing this line is because marge's commits were already pushed and these contained the `Reviewed-by` tags and it is extremely likely that these commits are the ones merged. For this not to happen, as far as I can tell, the following events should take place:

  1. Marge has pushed the commits and is waiting for CI to pass
  1. Someone pushes with `--force` new commits, say, without the tags
  1. GitLab is configured not to stop the ongoing pipelines when a new version of the branch is pushed (otherwise Marge would notice the pipeline was cancelled)
  1. The CI pipeline for the new version of the branch finishes faster than Marge's

I'd optimise for the common case of someone pressing "Merge when build succeeds" since, missing `Reviewed-by` tags should be caught by a git-hook anyway!

marge/job.py
jcpetruzza: we could leave a message in this case

aschmolck: I mean `log.warn` so when someone says "marge merged this funny"  we can look at marge.log and say, "ha,, not so fast, marge didn't merge this". We can still say this even if it's not logged, because the committer won't be marge, so it's just a detail.

jcpetruzza: Will add

PR #49, opened 2017-09-05
Fix check for the reapprove merge job option

benjamb:
I assume there may have been some confusion with the option as parsed with argparse and the attributes as part of MergeJobOptions.


jcpetruzza:
Good catch and thanks for the fix!

jcpetruzza:
Made a 0.3.1 release with your fix. It is already on docker-hub.

benjamb:
Thanks!

PR #50, opened 2017-09-22
URL encode branch names

benjamb:
Otherwise this will error if the branch contains a '/'.

### DIFF
--- a/marge/commit.py
+++ b/marge/commit.py
@@ -1,5 +1,7 @@
 import re
 
+from requests.utils import quote
+
 from . import gitlab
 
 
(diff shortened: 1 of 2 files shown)


benjamb:
Note in the API wrt branch/tag name encoding: https://docs.gitlab.com/ee/api/README.html#branches-amp-tags-name-encoding

jcpetruzza:
Good catch! Could you maybe add also a test-case for this in `test_commit.py`?

benjamb:
@jcpetruzza Will that do? Or would you prefer I adapt the existing test to cover this case?

jcpetruzza:
@benjamb It was perfect that way. Will make a new release, hopefully, later today

jcpetruzza:
Version 0.3.2 released

PR #51, opened 2017-10-04
Catch exception from early attempt to parse json in response

benjamb:
Currently attempts to log response.json() before verifying response status.

### DIFF
--- a/marge/gitlab.py
+++ b/marge/gitlab.py
@@ -19,7 +19,10 @@ class Api(object):
         log.debug('REQUEST: %s %s %r %r', method.__name__.upper(), url, headers, command.call_args)
         response = method(url, headers=headers, **command.call_args)
         log.debug('RESPONSE CODE: %s', response.status_code)
-        log.debug('RESPONSE BODY: %r', response.json())
+        try:
+            log.debug('RESPONSE BODY: %r', response.json())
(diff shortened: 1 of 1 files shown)


benjamb:
Wasn't certain whether or not I should just remove the debug statement?

aschmolck:
Did you encounter an API response that wasn't valid json in practice, and when did it happen? 

I think if this happens in practice,  we probably still want the to debug log the response body. Maybe the simplest solution wold be to just always print response.text (or if we feel paranoid, response.content) instead.

benjamb:
@aschmolck Yes. We've been hitting 503 errors. I'm not sure of the content of the response, only that an exception was thrown at this point.

Further down, when raising an exception against unrecognised response codes, the same try/except block is used. Should I update this patch to simply change `response.json()` to `response.text`?

aschmolck:
@benjamb you mean this bit, right?
```
        try:
            err_message = response.json()
        except json.JSONDecodeError:
            err_message = response.reason
```
Yeah, I think replacing both instances by `response.text` and getting rid of the try/except should probably be fine.

benjamb:
@aschmolck Cheers for sorting this out!

aschmolck:
@benjamb no problem: thanks for reporting bugs and submitting PRs!

Issue #54, opened 2017-10-09
Make git pull timeout configurable


siebertm:
in [marge/git.py](https://github.com/smarkets/marge-bot/blob/master/marge/git.py#L12), you're hardcoding the git clone timeout to 120 seconds. While this may be a sensible default, it absolutely doesn't work for large repositories which take longer to clone.

I'd rather propose this to be configurable by either an environment variable (`MARGE_GIT_TIMEOUT`?) or a command option (or both?)

glensc:
@siebertm never link to branch, use git tag or commit. `y` hotkey in github.

aschmolck:
Fixed in 0.4.0

PR #58, opened 2017-10-15
Add a --git-timeout parameter (that takes units)

aschmolck:
Also change --max-ci-waititing time to --ci-timeout  to work likewise.

### DIFF
--- a/marge/app.py
+++ b/marge/app.py
@@ -110,6 +110,12 @@ def _parse_args(args):
         help='How long to wait for CI to pass (defaut: 15min).',
     )
     arg(
+        '--max-ci-time-in-minutes',
+        type=int,
+        default=None,
+        help='Deprecated; use --ci-timeout.',
+    )
+    arg(
         '--git-timeout',
         type=time_interval,
         default=time_interval('120s'),
@@ -154,6 +160,12 @@ def main(args=sys.argv[1:]):
     with _secret_auth_token_and_ssh_key(options) as (auth_token, ssh_key_file):
         api = gitlab.Api(options.gitlab_url, auth_token)
         user = user_module.User.myself(api)
+        if options.max_ci_time_in_minutes:
+            logging.warning(
+                "--max-ci-time-in-minutes is DEPRECATED, use --ci-timeout %dmin",
+                options.max_ci_time_in_minutes
+            )
+            options.ci_timeout = timedelta(minutes=options.max_ci_time_in_minutes)
 
         config = bot.BotConfig(
             user=user,
--- a/marge/app.py
+++ b/marge/app.py
@@ -17,6 +17,16 @@ from . import gitlab
 from . import user as user_module
 
 
+
+def time_interval(s):
+    try:
+        quant, unit = re.match(r'\A([\d.]+) ?(h|m(?:in)?|s)?\Z', s).groups()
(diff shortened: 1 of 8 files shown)


marge/app.py
```diff
@@ -17,6 +17,16 @@
...
 
+
+def time_interval(s):
+    try:
+        quant, unit = re.match(r'\A([\d.]+)(h|m(?:in)?|s)?\Z', s).groups()
+        translate = {'h': 'hours', 'm': 'minutes', 'min': 'minutes', 's': 'seconds'}
+        return timedelta(**{translate[unit or 's']: float(quant)})
+    except (AttributeError, ValueError):
+        raise argparse.ArgumentTypeError('Invalid time interval (e.g. 12[s|min|h]): %s', s)
+
```
jcpetruzza: @aschmolck Could you also allow for a space between number and unit? E.g. `2 min`. Also, since you allow for `m` and `min`, how about allowing for `sec` as well?

marge/app.py
```diff
@@ -94,10 +104,16 @@ def regexp(s):
...
-        '--max-ci-time-in-minutes',
-        type=int,
-        default=15,
-        help='How long to wait for CI to pass.',
+        '--ci-timeout',
+        type=time_interval,
+        default=time_interval('15min'),
+        help='How long to wait for CI to pass (defaut: 15min).',
+    )
+    arg(
```
jcpetruzza: @aschmolck I understand why you want to be consistent here, but simply dropping the old flag is unnecessarily harsh on users: most of them will run it directly from docker, probably using the `latest` tag, so next time it is restarted, a new version will be pulled, and suddenly their installation will not work anymore, it will take time to figure out why, etc. 

I would suggest keeping the old flag as well, maybe printing something along the lines of "this option is deprecated, use --ci-timeout instead".

marge/app.py
aschmolck: The only reason I allow m(eters) for minutes is because that's unfortunately some sort of dumb pseudo standard (IIRC, grafana, prometheus, gnu sleep etc. all use `m` rather than `min`). Optional space, OTOH, I'm fine with – pushed.

marge/app.py
aschmolck: I thought p(does_not_use_defaul)*user_base was probably small enough, but you're right it's nicer to err on site of caution.

aschmolck:
This closes #54.

marge/app.py
```diff
@@ -154,6 +160,12 @@ def main(args=sys.argv[1:]):
     with _secret_auth_token_and_ssh_key(options) as (auth_token, ssh_key_file):
         api = gitlab.Api(options.gitlab_url, auth_token)
         user = user_module.User.myself(api)
+        if options.max_ci_time_in_minutes:
+            logging.warning(
+                "--max-ci-time-in-minutes is DEPRECATED, use --ci-timeout %dmin",
+                options.max_ci_time_in_minutes
+            )
+            options.ci_timeout = timedelta(minutes=options.max_ci_time_in_minutes)
```
jcpetruzza: @aschmolck I rebased my branch with unit tests for the `app` on top of this and it caught the error here! It should be:
```python
 options.ci_timeout = timedelta(minutes=options.max_ci_time_in_minutes).total_seconds()
```

Don't make a release until I push that branch with the fix...

PR #59, opened 2017-10-15
Closes #53: reapprove immediately after push

### LINKED ISSUE #53: Re-approvals only applied after successful CI run
There is a comment related to this within `marge/job.py`, including for context:
```
# Re-approve the merge request, in case us pushing it has removed
# approvals. Note that there is a bit of a race; effectively
# approval can't be withdrawn after we've pushed (resetting
# approvals) and CI runs.
```

Occasionally CI may fail due to transient network issues that are unrelated to the change made. In this case, Marge will error out and not bother attempting to reapply any approvals. GitLab doesn't remove approvals on CI failure, so it doesn't quite make sense that this happens with Marge.

This also applies to any potential exception that might occur between the force push and applying approvals, we need to restart marge and then manually approve again.

I'm unaware as to whether there is a historical reason for why approvals are reapplied when they are, but could they no be applied immediately after the rebase?

aschmolck:
Rather than waiting for CI to pass, just approve right away; this is more
convenient for transient errors such as flaky builds, and I see no obvious
downside.

### DIFF
--- a/marge/job.py
+++ b/marge/job.py
@@ -63,7 +63,7 @@ class MergeJob(object):
             log.exception('Unexpected Git error')
             merge_request.comment('Something seems broken on my local git repo; check my logs!')
             raise
-        except Exception:
+        except Exception as _ex:
             log.exception('Unexpected Exception')
             merge_request.comment("I'm broken on the inside, please somebody fix me... :cry:")
             self.unassign_from_mr(merge_request)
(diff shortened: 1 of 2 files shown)


marge/job.py
```diff
@@ -63,7 +63,7 @@ def execute(self):
             log.exception('Unexpected Git error')
             merge_request.comment('Something seems broken on my local git repo; check my logs!')
             raise
-        except Exception:
+        except Exception as _ex:
```
jcpetruzza: @aschmolk Why this change?

tests/test_job.py
```diff
@@ -90,8 +90,13 @@ def __init__(self, gitlab_url=None):
...
         api.add_transition(
            GET('/projects/1234/repository/branches/useless_new_feature'),
+           Ok({'commit': _commit(id=rewritten_sha, status='running')}),
+            from_state='pushed',
+        )
+        api.add_transition(
+           GET('/projects/1234/repository/branches/useless_new_feature'),
            Ok({'commit': _commit(id=rewritten_sha, status='success')}),
-           from_state='passed'
+            from_state='passed'
```
jcpetruzza: @aschmolck Extra space before `from_state` (also in the transition above)

marge/job.py
aschmolck: Just for debugging, it's convenient to have the exception around.

tests/test_job.py
aschmolck: Weird, that was due to a bug in emacs python mode indentation logic!

PR #61, opened 2017-10-18
Always use --ssh-key-file, if passed

aschmolck:
Ssh needs some irritating contortions to make this happen, see comment in commit.

### DIFF
--- a/marge/git.py
+++ b/marge/git.py
@@ -126,7 +126,17 @@ class Repo(namedtuple('Repo', 'remote_url local_path ssh_key_file')):
         env = None
         if self.ssh_key_file:
             env = os.environ.copy()
-            env['GIT_SSH_COMMAND'] = " ".join([GIT_SSH_COMMAND, "-i", self.ssh_key_file])
+            # ssh's handling of identity files is infuriatingly dumb, to get it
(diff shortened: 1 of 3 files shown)


marge/git.py
```diff
@@ -126,7 +126,17 @@ def git(self, *args, from_repo=True):
...
+            # ssh's handling of identity files is infuriatingly dumb, to get it
+            # to actualy really use the IdentityFile we pass in via -i we also
+            # need to tell it to ignore ssh-agent (IdentitiesOnly=true) and not
+            # read in any identities from ~/.ssh/config etc (-F /dev/null),
+            # because they append and it tries them in order, starting with config file
+            env['GIT_SSH_COMMAND'] = " ".join([
+                GIT_SSH_COMMAND,
+                "-F", "/dev/null",
+                "-o", "IdentitiesOnly=yes",
+                "-i", self.ssh_key_file
```
jcpetruzza: @aschmolck style: missing trailing comma...

jcpetruzza:
I think this is only an issue when you run marge-bot locally from your box (and not inside a docker container), but It is a good change in any case. Once the unit test is fixed, feel free to merge

PR #62, opened 2017-10-18
Now that we got control of smarkets on docker hub use that...

aschmolck:
... not smarketshq.

### DIFF
--- a/dockerize.nix
+++ b/dockerize.nix
@@ -5,7 +5,7 @@ let callPackage = pkgs.lib.callPackageWith (pkgs);
   version = marge.version;
 in
 dockerTools.buildImage {
-  name = "smarketshq/marge-bot";
+  name = "smarkets/marge-bot";
   tag = "${version}";
   # minimal user setup, so ssh won't whine 'No user exists for uid 0'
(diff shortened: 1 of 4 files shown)


dockerize.nix
```diff
@@ -5,7 +5,7 @@ let callPackage = pkgs.lib.callPackageWith (pkgs);
   version = marge.version;
 in
 dockerTools.buildImage {
-  name = "smarketshq/marge-bot";
+  name = "smarkets/marge-bot";
```
jcpetruzza: @aschmolck I'd add a `repo` parameter that defaults to `"smarkets"` and here I'd use:
```nix
name = "${repo}/marge-bot";
```

Then we can make the `Makefile` produce and push both to both places, at least for now.

aschmolck: It's not really much harder to push manually to both, and I think we only want to do this for one or two releases.

jcpetruzza: That will lead to inconsistent releases where some are pushed to one, some to both, etc. Until we decide to actually terminate the old repository, I'd prefer if our scripts push to both.

PR #65, opened 2017-11-16
Add feature to restrict marge to a specific branch

benjamb:
Copies the same debug logging as `--project-regexp` when filtering. It makes use of the MRs web_url as it's easily identifiable.

This has been quite useful for the project I'm working on, since not only can CI last a long time, but we currently have two branches that are targetted for a high number of MRs. This has allowed us to run an instance of Marge for both of these branches, allowing for some fast and dirty parallelisation.


jcpetruzza:
@benjamb This is cool, thanks for submitting it!

PR #66, opened 2017-11-16
Reduce logging in non-debug

benjamb:
Logs can very quickly get swamped with new HTTPS connection from requests and CI polling messages. I understand that in most use cases, CI likely won't last too long, but these two messages take up the majority of the logs.

### DIFF
--- a/marge/job.py
+++ b/marge/job.py
@@ -203,6 +203,7 @@ class MergeJob(object):
         time_0 = datetime.utcnow()
         waiting_time_in_secs = 10
 
+        log.info('Waiting for CI to pass')
         while datetime.utcnow() - time_0 < self._options.ci_timeout:
             ci_status = Commit.fetch_by_id(source_project_id, commit_sha, api).status
             if ci_status == 'success':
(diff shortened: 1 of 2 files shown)


marge/job.py
```diff
@@ -217,7 +217,7 @@ def wait_for_ci_to_pass(self, source_project_id, commit_sha):
             if ci_status not in ('pending', 'running'):
                 log.warning('Suspicious build status: %r', ci_status)
 
-            log.info('Waiting for %s secs before polling CI status again', waiting_time_in_secs)
+            log.debug('Waiting for %s secs before polling CI status again', waiting_time_in_secs)
```
jcpetruzza: @benjamb This is a good idea, but we should probably add something like `log.info("Waiting for CI to pass...)` before the while loop, so we can check the logs and easily tell what marge is doing.

benjamb: @jcpetruzza You're right. I'll update shortly.

benjamb: Done!

jcpetruzza: Excellent, thanks!

PR #70, opened 2017-12-01
Auto release

mrusu91:


Makefile
```diff
@@ -17,9 +17,9 @@ dockerize: dockerize.nix
 	docker load --input $$(nix-build dockerize.nix)
 
 
-.PHONY: dockerhub
+.PHONY: docker-push
 docker-push:
-	docker login
+	docker login -u "$${DOCKER_USERNAME}" -p "$${DOCKER_PASSWORD}"
```
aschmolck: I"d prefer something like 
```
if [[ -n $$DOCKER_USERNAME && -n $$DOCKER_PASSWORD ]]; then \ 
  docker login -u "$${DOCKER_USERNAME}" -p "$${DOCKER_PASSWORD}" \ 
else \ 
 docker login  \
fi 
```

marge.nix
```diff
@@ -8,22 +7,12 @@ python.mkDerivation {
   version = "${version}";
   name = "marge-${version}";
   src = ./.;
-  # The dependencies, referring to variables in <nixpkgs>.
-  buildInputs = [py.pylint py.pytest py.pytest-cov];
+  buildInputs = [py.pytest py.pytest-cov py.pytest-pylint py.pytest-runner];
   propagatedBuildInputs = [py.maya py.requests pkgs.openssh pkgs.git];
-  checkPhase = ''
```
aschmolck: Hmm, what happened to tests?

aschmolck: Never mind, py.pytest-runner. Wonder if we need the PYTHONDONTWRITEBYTECODE.

mrusu91: I think is fine

tests/test_git.py
```diff
@@ -234,7 +234,7 @@ def test_filter():
 
 def test_filter_fails_on_empty_commit_messages():
     with pytest.raises(Exception):
-        _filter_test('', [])
+        _filter_test('', '', [])
```
jcpetruzza: @mrusu91 This was a good catch. Now, the question is why wouldn't this fail when running the unit-tests? And the answer is: "because we are using `pytest.raises(Exception)`, which is too general, and was catching the exception raised by `_filter_test()` being called with the wrong arguments.

I think we should instead replace it with:
```python
 def test_filter_fails_on_empty_commit_messages():
      with pytest.raises(subprocess.CalledProcessError):
          _filter_test('', '', [])
```
and similarly with the other test cases.

jcpetruzza: I fixed this in 1438867, where in fact I check that the error message we get is what we were expecting (or else, the script could be crashing instead). Interestingly, there was a bug in the error handling of `trailfilter` that only surfaced when I tried to do this...

PR #71, opened 2017-12-04
Add config file -> env var -> args configuration options

mrusu91:
This commit makes it easiser to configure marge-bot when it's deployed
in environments where cli args or ENV vars are not feasible

Implements the following rule:
"default -> config file -> env var -> args"
where right overwrites everyting on the left side

This commit does the following:

1. Switches from 'argparse' to 'configargparse' lib because it makes it
easy to implement the rule
2. Adds MARGE_SSH_KEY_FILE and MARGE_AUTH_TOKEN_FILE as ENV vars to
comply with the rule
3. Adds '--auth-token' and '--ssh-key' as cli args to comply with the
rule but raise an exception if set via args
4. Makes ('--ssh-key' and '--ssh-key-file') and ('--auth-key' and
'--auth-key-file') mutually exclusive

Signed-off-by: Marian Rusu <rusumarian91@gmail.com>


marge/app.py
```diff
@@ -27,138 +28,153 @@ def time_interval(s):
...
+    auth_token_group.add_argument(
         '--auth-token-file',
         type=argparse.FileType('rt'),
         metavar='FILE',
-        help='Your gitlab token; must provide this flag or set MARGE_AUTH_TOKEN',
-        env_var=False,
+        help='Path to your gitlab token file.\n',
     )
-    arg(
+    parser.add_argument(
```
aschmolck: I kinda prefer:
```
arg = parse.add_argument 
arg(...
```
but up to you.

marge/app.py
```diff
@@ -27,138 +28,153 @@ def time_interval(s):
...
+    )
+    auth_token_group = parser.add_mutually_exclusive_group(required=True)
+    auth_token_group.add_argument(
+        '--auth-token',
+        type=str,
+        metavar='TOKEN',
+        help=(
+            'Your gitlab token.\n'
+            'WARNING:\n'
+            'Due to security reasons, you can only set this via ENV var or config file.\n'
```
aschmolck: mention `--auth-token-file`?

aschmolck:
(i.e. feel free to merge)

marge/app.py
mrusu91: hmm, what's the reason?

marge/app.py
mrusu91: This is about '--auth-token', 'MARGE_AUTH_TOKEN' and 'auth-token' in config file. What i tried to say is use ENV var or config map.

I think '--auth-token-file' should be seen as a separate option?

aschmolck: But it is related. If you want to use a commandline flag (vs config file or env var), you should use --auth-token-file instead of --auth-token. 

I'd write somethihng like "DISABLED because passing credentials on the command line is insecure; use the identical config file option, set MARGE_AUTH_TOKEN or instead specify a file with --auth-token-file'

marge/app.py
aschmolck: Matter of taste. Less indent, more concise, more readable IMO and makes tweaking easier. I'm fine w/ you leaving it as is though.

marge/app.py
```diff
@@ -11,13 +11,14 @@
 import tempfile
 from datetime import timedelta
 
+import configargparse
```
jcpetruzza: @mrusu91 You are still importing `argparse` and it's being used (e.g. raising `argparse.ArgumentTypeError`). I'd prefer if we use only `configargparse` everywhere. Maybe we can just do:
```python
import configargparse as argparse
```

marge/app.py
```diff
@@ -27,138 +28,153 @@ def time_interval(s):
...
-            if env_var_val:
-                if kwargs.get('action') in ['store_true', 'store_false']:
-                    env_var_val = env_var_val.lower() in ['true', 'yes', '1']
-                kwargs = dict(kwargs, default=env_var_val, required=False)
+def _parse_config(args):
 
-        parser.add_argument(name, **kwargs)
+    for bad_arg in ['--auth-token', '--ssh-key']:
+        if bad_arg in args:
+            raise Exception('"%s" can only be set via ENV var or config file.' % bad_arg)
```
jcpetruzza: @mrusu91 I can still call marge with `--auth-token=everybody-can-see-my-token` and this test won't catch it, right?

marge/app.py
```diff
@@ -27,138 +28,153 @@ def time_interval(s):
...
+        help='Add "Tested: marge-bot <$MR_URL>" for the final commit on branch after it passed CI.\n',
     )
-    arg(
+    parser.add_argument(
         '--add-part-of',
         action='store_true',
-        help='add Part-of: <$PR_URL> to each commit in PR (except top commit when --add-tested is used)',
+        help=(
+            'Add "Part-of: <$MR_URL>" to each commit in MR\n'
+            '(except top commit when --add-tested is used).\n'
```
jcpetruzza: @mrusu91 Actually, I think the `help` entry here was wrong: we are adding `Part-of: ` on every commit, regardless of whether `--add-tested` is used.

marge/app.py
jcpetruzza: Also, can we maybe add a test for this case in `test_app.py`?

jcpetruzza:
@mrusu91 So, if I understood correctly, the config file will only be used if `--config-file=marge.yaml` is passed as a command-line arg, right? Out of curiosity, will `MARGE_CONFIG_FILE` work as well?

It would be nice to have at least one test-case showing that it reads stuff from the config file and the config > env > arg rule being satisfied.

But more importantly, we should probably update the README file to explain we how the config file is used and maybe give a short example of the configuration syntax. What do you think?

marge/app.py
mrusu91: will do that, but maybe keep using the original name `configargparse` because I think it keeps  things clear?

marge/app.py
mrusu91: hmm, that's true, will find  a solution and write a test

marge/app.py
mrusu91: so should are remove the 2nd line?

mrusu91:
Good points, 

`MARGE_CONFIG_FILE` doesn't work now but i think i can fix it. 

The test case here sounds like it's out of scope of marge. The test should be in `configargparse` project?
see: https://github.com/bw2/ConfigArgParse/blob/master/tests/test_configargparse.py

I will update the README as well

marge/app.py
jcpetruzza: :+1:

marge/app.py
jcpetruzza: yes, please!

jcpetruzza:
I think I'd at least have a test that `--config-file` works. And once you have such a test, making it so some config file parameters are overridden by env/cmdline is almost free. We are already doing this with env/cmdline alone, you just need to add a new context-manager for config-files.

marge/app.py
```diff
@@ -11,154 +10,177 @@
...
+        '--debug',
+        action='store_true',
+        help='Debug logging (includes all HTTP requests etc).\n',
     )
-    arg('--debug', action='store_true', help='Debug logging (includes all HTTP requests etc.)')
+    config = parser.parse_args(args)
 
-    return parser.parse_args(args)
+    cli_args = []
+    for _, (_, value) in parser._source_to_settings.get(configargparse._COMMAND_LINE_SOURCE_KEY, {}).items():
```
jcpetruzza: a bit ugly, but ok :smile:

tests/test_app.py
```diff
@@ -144,3 +170,58 @@ def test_branch_regexp():
 def test_time_interval():
     _900s = datetime.timedelta(0, 900)
     assert [app.time_interval(x) for x in ['15min', '15m', '.25h', '900s']] == [_900s] * 4
+
+
+def test_disabled_auth_token_cli_arg():
+    with env(MARGE_SSH_KEY="KEY", MARGE_GITLAB_URL='http://foo.com'):
+        with pytest.raises(Exception):
```
jcpetruzza: @mrusu91 Could you pass a more specific class here?

PR #78, opened 2018-02-02
Prune stale branches when fetching from remote

benjamb:
Fixes occasional error where local refs conflict with changes on the
remote:

    error: some local refs could not be updated; try running
    'git remote prune origin' to remove any old, conflicting branches


aschmolck:
Thanks for reporting the issue and sending an PR to fix it!

But I think instead of the more esoteric remote update --prune, we could just stick with fetch and add `--prune`, correct?

benjamb:
Yep, correct. Unsure why I didn't just resort to that, probably because I tend to use `git remote update --prune` rather than fetch or pull myself. Force pushed an update, it now sticks with fetch.

jcpetruzza:
Thanks for the fix!

PR #85, opened 2018-02-21
Wait for approvals race

benjamb:
Add option to wait for approvals to reset after pushing.

When using the "new commits remove all approvals" option in a project's settings, it is possible to run into a race condition post-rebase where the approvals haven't reset by the time we check. This results in the later attempt to merge reporting that there are not enough approvals.

### DIFF
--- a/marge/app.py
+++ b/marge/app.py
@@ -129,6 +129,17 @@ def _parse_config(args):
         help='Marge-bot pushes effectively don\'t change approval status.\n',
     )
     parser.add_argument(
+        '--approval-reset-timeout',
+        type=time_interval,
+        default='0s',
+        help=(
+            'How long to wait for approvals to reset after pushing.\n'
+            'Only useful with the "new commits remove all approvals" option in a project\'s settings.\n'
+            'This is to handle the potential race condition where approvals don\'t reset in GitLab\n'
+            'after a force push due to slow processing of the event.\n'
+        ),
+    )
+    parser.add_argument(
         '--project-regexp',
         type=regexp,
         default='.*',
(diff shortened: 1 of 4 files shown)


aschmolck:
Thanks for this PR, this is definitely something we should fix!

I'm leaning to not making this an option and instead always activate the logic if the project has `new commits remove all approvals set` -- do you think it would be possible to change it that way?

It seems like you probably always want this race condition work around if you have push-approval-removal activated in gitlab, so rather than adding yet one more option and have people discover the hard way that they need to activate it if they use a certain setting, why not let marge-bot figure it out herself?

benjamb:
@aschmolck I probably should have mentioned. I don't believe this is something that can be determined from the API, at least, it's not exposed as anything I can change here: https://docs.gitlab.com/ce/api/projects.html#edit-project

Hence why I implemented that as an option. I would absolutely prefer that marge-bot be able to discover this herself though.

aschmolck:
Yeah, I didn't see something in the API description either (note that your link is to CE and not EE though: I think we want https://docs.gitlab.com/ee/api/projects.html#edit-project).

One would think if it's to be found somewhere it would be either at the projects/ endpoint or the approvals one (`/projects/{0.project_id}/merge_requests/{0.iid}/approvals`). Could you have a look at the json that gets returned for both the project and approvals endpoint and see if anything is in there (for a project that has reset-on-push activated)?

Otherwise we should probably merge this as an experimental feature and file a bug with gitlab to either expose the information so we can detect the work-around is needed or keep the resetting timeframe reasonably short; if it's within say 5s, we can just always put an unconditional sleep 5s there.

benjamb:
I've checked both endpoints and have not found anything suitable unfortunately. It appears it will be implemented in the not so distant future as part of: https://gitlab.com/gitlab-org/gitlab-ee/issues/183

I'm unsure as to whether this would be backported to earlier versions though.

aschmolck:
Thanks -- for the research, I've merged this now as is. Should probably add something to the README before the next release.

PR #87, opened 2018-03-01
Fail to merge if changes already exist in target

JaimeLennox:
It's possible that the changes in an MR already exist in the target branch - whether this was merged in manually or someone else happened to have a similar idea. If this happens, when we rebase the MR on the target the commits will be dropped, essentially turning into a no-op. This causes the filter-branch to fail as it expects some revisions to be listed. Whilst this has the same end result, it's clearer if we explain exactly what the issue is.

### DIFF
--- a/marge/job.py
+++ b/marge/job.py
@@ -312,12 +312,15 @@ def update_from_target_branch_and_push(
     branch_updated = branch_rewritten = changes_pushed = False
     try:
         fuse = repo.merge if use_merge_strategy else repo.rebase
+        target_sha = repo.get_commit_hash('origin/' + target_branch)
         rewritten_sha = updated_sha = fuse(
             branch=source_branch,
             new_base=target_branch,
             source_repo_url=source_repo_url
         )
         branch_updated = True
+        if updated_sha == target_sha:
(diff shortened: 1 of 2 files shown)


jcpetruzza:
@JaimeLennox IIUC, the fix will only detect the problem when the MR was "just" merged and is already on top of the target branch. But other merges could have happened  in the meantime, or someone could have accidentally merged in the commits as part of another branch.

If I were doing marge's job manually, I'd notice that after rebasing the branch locally, what I got in the end was the target branch HEAD (i.e., no commit was applied on top of the new base) and at that point I could realise that there is nothing merge. Maybe this could be implemented in the same way?

JaimeLennox:
Unless I'm misunderstanding what you're saying, I'm already doing it that way? After we rebase, we check the head of the branch and compare it to the target branch commit, and if they're the same then we raise the error.

jcpetruzza:
Yeah, you are right, I misread the code flow!

jcpetruzza:
Looks good to me. Did you consider adding a test case to `test_job` for this? The `repo` is already a mock there, so you could make it so `repo.merge()` returns the target-branch hash and check you get the right exception.

JaimeLennox:
Sure - I've added a test.

PR #96, opened 2018-03-23
Get latest CI status of a commit filtered by branch

benjamb:
The current implementation fetches the latest CI status for a particular sha. While a relatively rare circumstance, it's possible that at the point Marge-bot picks up an MR, another developer may have branched from the same commit and pushed to the repository. This new pipeline would be detected by Marge-bot, which could then be cancelled, causing Marge-bot to drop the MR stating "CI cancelled", while the CI for the correct branch will have passed.

This MR changes the CI status check to use the pipelines API, in order to filter status by ref (branch).

Commit.fetch_by_id() is no longer used after this change, I wasn't entirely sure whether I should remove that as part of this, leave it for a follow-up, or for you to decide what to do with it.

### DIFF
--- a/marge/job.py
+++ b/marge/job.py
@@ -9,6 +9,7 @@ from .commit import Commit
 from .interval import IntervalUnion
 from .project import Project
 from .user import User
+from .pipeline import Pipeline
 
 
 class MergeJob(object):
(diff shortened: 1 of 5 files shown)


JaimeLennox:
Looks good to me. Thanks!

PR #108, opened 2018-05-21
Remove bare asterisk syntax

guoqiao:
Remove trailing comma in function arg list with bare asterisk
    
    In Python 3.x before 3.6, there is a bug:
    
    It allows a trailing comma after arg list:
    
        def f(a, b,):
            pass
    
    But raise invalid syntax error when you have bare asterisk arg:
    
        def f(*, a, b,)
            pass
    
    Once we make this change, marge-bot code is ok to run with other python
    3 versions.
    
    Refer to: https://bugs.python.org/issue9232

### DIFF
--- a/marge/job.py
+++ b/marge/job.py
@@ -284,7 +284,7 @@ def update_from_target_branch_and_push(
         reviewers=None,
         tested_by=None,
         part_of=None,
-        use_merge_strategy=False,
+        use_merge_strategy=False
 ):
     """Updates `target_branch` with commits from `source_branch`, optionally add trailers and push.
     The update strategy can either be rebase or merge. The default is rebase.


guoqiao:
@aschmolck  I did get the invalid syntax error for Python 3.5, but the reason I gave was wrong. I figured out the root reason, please refer to my modified comment above.

Jellby:
@guoqiao Apparently there was some recent push with older commits that causes a conflict. I have the same problem with my MRs, and only noticed because I checked them, I didn't get any notification, so if you didn't either, now you know...

JaimeLennox:
I'm not really sure this is an issue - we provide a Docker image and nix environment where your Python version shouldn't matter, otherwise you can create a virtualenv with Python 3.6.

Issue #107, opened 2018-05-21
Trigger pipelines for merge requests


Jellby:
Since GitLab does not support pipeline for merge requests yet (see https://gitlab.com/gitlab-org/gitlab-ce/issues/23902), can marge-bot take care of it?

I'd like her to:
1. Check if a pipeline has been created on the MR (or the branch it comes from)
2. Check if it is the correct pipeline (if it includes the right jobs, at least)
3. If either is no, trigger a pipeline for the MR and wait for completion

The reason is that for expensive tests one may not want to run them for every push to every branch, but they must be run on a MR if it's going to go into `master`. The CI configuration could be set up to run these tests only for, say, `triggers` and `web`, but can't enforce them to run on MRs. A possibility is running the test for some specially-named branches by setting a branch regexp for the CI, and require MRs to use these names, but that needs some discipline which is hard to find (although marge-bot could help here too).

Jellby:
The tricky part would be detecting if a MR pipeline has already been triggered. In case of misconfiguration, a trigger may fail to create the desired jobs, in which case a naïve implementation (trigger if jobs do not exist, wait and repeat) would result in infinite triggers. The origin of a pipeline is passed as a CI variable (`CI_PIPELINE_SOURCE`) but it does not seem to be available in the API. Maybe Marge could leave a message in the discussion saying "I triggered a pipeline for [sha]", and check existing messages before triggering a pipeline? Or maybe it's just better to use some other service to trigger pipelines for MRs until it's properly implemented in GitLab (see https://github.com/shyiko/gitlab-ci-build-on-merge-request).

snim2:
This has been overtaken by the new [rules for GitLab CI](https://docs.gitlab.com/ee/ci/yaml/#rules) which includes support for MR pipelines, so maybe needs a re-think?

Jellby:
Triggering pipelines for MRs is no longer an issue (for me). But working with forks, or ensuring the pipeline is of the right type still is.

snim2:
That's very useful, thanks. I work with forks so rarely I forget that they are quite different in GitLab.

PR #109, opened 2018-05-23
Add create-pipeline and job-regexp options

Jellby:
This addresses #107 

`create-pipeline`: If for some reason no valid pipeline is found for the MR (none exists, outdated, etc.), Marge will create a new one using the api. She will also leave a message, so she can know that she already created a pipeline and will not do it again and again. Default is false.

`job-regexp`: The CI may be configured to run different jobs depending on the branch name, or pipeline source (push, trigger, api...). Marge will check the jobs in the current pipeline against this regexp. If any job matches the regexp, it's considered good and the status is checked as usual. If none matches, it's not a valid pipeline and either the merge will fail or a new pipeline will be created (see `create-pipeline`).

If using both `create-pipeline` and `job-regexp`, make sure the important jobs are configured to run on `api` sources (see the `only` and `except` options of GitLab CI), so that the pipeline created by Marge will be valid.

### DIFF
--- a/marge/app.py
+++ b/marge/app.py
@@ -170,6 +170,17 @@ def _parse_config(args):
         help='Only process MRs whose target branches match the given regular expression.\n',
     )
     parser.add_argument(
+        '--job-regexp',
+        type=regexp,
+        default='.*',
+        help='Require pipelines to have jobs matching the given regular expression.\n',
+    )
+    parser.add_argument(
+        '--create-pipeline',
+        action='store_true',
+        help='Create new pipeline if not up to date or not matching job-regexp.\n',
+    )
+    parser.add_argument(
         '--debug',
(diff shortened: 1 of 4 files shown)


Jellby:
I'm trying to understand how `gitlab_api_mock.py` works, because it seems I'd have to implement something for the jobs API (I could use some help). But is it worth continuing?

jcpetruzza:
In case it helps, you use `gitlab_api_mock.py` to define a mock of gitlab, expressed as a state machine; depending in which state you are, you expect certain requests and respond in certain way, and on certain requests you may transition to a different state. There is some mechanism to say that some requests are expected on every state, etc. Check the definition of `MockLab` inside `test_job.py` to see an example of how it works.

Issue #104, opened 2018-05-19
Crash if no pipeline is run


Jellby:
I was testing what happens if no pipeline is run for the MR. Here's what I got (using merge strategy, if it matters):

```
2018-05-19 15:19:53,666 INFO Commit id to merge '8a8a087a54ea54e464af4dcf2bd457933b00dbe7' (into: '946a97b54cbcd4a47f21a1bed828e87c715a401b')
2018-05-19 15:19:59,405 INFO Waiting for CI to pass
2018-05-19 15:20:00,419 ERROR Unexpected Exception
Traceback (most recent call last):
  File "/home/ubuntu/marge-bot/marge/job.py", line 59, in execute
    self.update_merge_request_and_accept(approvals)
  File "/home/ubuntu/marge-bot/marge/job.py", line 152, in update_merge_request_and_accept
    self.wait_for_ci_to_pass(source_project.id, actual_sha, merge_request.source_branch)
  File "/home/ubuntu/marge-bot/marge/job.py", line 224, in wait_for_ci_to_pass
    assert current_pipeline.sha == commit_sha
AssertionError
```

I guess this is because there's no pipeline at all (configured with `only: master`, and `master` is the *target* branch, not the source). A graceful handling and a post in the MR would have been much better.

aschmolck:
This looks like a bug. Note that `--use-merge-strategy` is experimental (and marked as such), and generally not super well behaved, we accepted an MR to add minimal support, but it's not very good yet and I just accepted another one (belatedly) that fixed a bad function call (https://github.com/smarkets/marge-bot/pull/102, also see https://github.com/smarkets/marge-bot/issues/72).

Meta question: Irrespective of its current state of implementation, are you sure you want to --use-merge-strategy in the first place? Note that you can still can have your source branches merged into master (rather than rebased) by just leaving the default MR strategy in gitlab active (which does that). You only need this options if it's important to you that updating the source branch to the latest commit in the target branch happens via merge rather than rebase (and I'm still not sure why one would want that).

Jellby:
I guess the crash happens when a pipeline was run for the branch, but not for its latest commit (`assert current_pipeline.sha == commit_sha`), which can happen if pipelines are run manually or if the CI configuration changes. And the problem is having an `assert` with no exception handling.

[...]

PR #112, opened 2018-06-04
Do not push to source repository if there was no change

Jellby:
In the simplest case where the MR can be immediately merged, there's no need for Marge to push to the source repository. In the case of a MR originating from a fork, Marge could have no permission to push, and then the merge fails for no good reason.

Proper handling of MRs from forks will come in a different PR (#113), but this looks like a harmless change that could be accepted right away.

### DIFF
--- a/marge/job.py
+++ b/marge/job.py
@@ -255,7 +255,11 @@ class MergeJob(object):
                 raise CannotMerge('these changes already exist in branch `{}`'.format(target_branch))
             rewritten_sha = self.add_trailers(merge_request) or updated_sha
             branch_rewritten = True
-            source_sha = repo.get_commit_hash('source/' + source_branch)
+            if merge_request.source_project_id == self._project.id:
+                source_name = 'origin/'
+            else:
+                source_name = 'source/'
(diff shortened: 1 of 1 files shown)


aschmolck:
I don't think adding extra logic to special case this is a good idea, because now rather than sorting out your permissions you have a situation where sometimes things fail after they appeared to work fine before. 

I'd much rather keep it so that if you want to use marge, it needs to be able to push to your branch.

Jellby:
With the merge strategy there's no need to push to source branches (for reasons to use the merge strategy, see https://github.com/smarkets/marge-bot/issues/104#issuecomment-390641138). Would it be acceptable if the check is done on the merge strategy instead?

Jellby:
@aschmolck With the merge strategy there's no reason for marge to push to other branches. Would it be acceptable to skip the push when using the merge strategy?

Even with the rebase strategy I don't see a problem with pushing only when necessary. If marge has no permission to push and it's not needed, the merge can proceed anyway. If a push is needed, she will leave a message. Is that worse than leaving a message and failing no matter what, even if no push is needed?

Issue #111, opened 2018-05-30
Handling merge requests from forks


Jellby:
A very significant flaw of the merge request model in GitLab is that it seems the only CI linked to the merge request is in the source project. This may be alright in very simple projects and configurations, but I don't think the CI from an external project should be trusted:

1. It could have no access to specific runners
2. Secret variables will not be available
3. Runner time limits could be different

These could cause CI jobs in a fork to fail, with almost no chance of fixing them. Perhaps more serious, I imagine it could be possible for a malicious user to set up a customized runner that will report success where a test should fail, in which case the CI would appear as passing, but it will be broken when run in the target project's runners.

I don't know if there's any way to run pipelines for merge requests on the target project. Apparently the  SHA for the MR commit is created in the target project, and a pipeline can be run if a tag or branch is added to it, but it is not connected to the MR. See for example:

https://gitlab.com/Jellby/test/merge_requests/10 The CI on the source project fails (artificially forced to do so, but it could happen in a real case for legitimate reasons as seen above).

The commit appeared in the target repository: https://gitlab.com/Jellby/test/commit/1ca482baa35df4287afaf5313a2f64fc59b62ed5

I created a tag: https://gitlab.com/Jellby/test/commits/testingMR!10 and ran a pipeline: https://gitlab.com/Jellby/test/pipelines/22898164 on the target project, but the MR does not show its result, so the MR can't be accepted.

If someone can find a good solution, that would be great. Otherwise, I'd suggest at least an option to ignore MRs not originating from the same project (and it should be the default if the guarantee for a green master is important).

Jellby:
Everything Marge does (or at least everything I want her to do) is an issue with GitLab, it should all be handled natively by GitLab. But I thought the point was to implement what GitLab was missing.

nanonyme:
Sure. We did bump into this I think three times over past 24 hours with new contributors.

RalfJung:
This is a roadblock for us. I originally thought marge-bot works like bors, where the bot pushes the to-be-pushed-to-master commit (MR + master) to a `staging` branch, tests that, and then fast-forwards master if CI succeeds. That would help with the fork situation as well since this new branch is located in the main repository, not the fork. But it looks like marge-bot actually relies on CI in the original MR branch, and that won't cut it for us as we need specific runners that forks do not have access to...

Jellby:
[...]

PR #113, opened 2018-06-04
Temporary branch for external MRs

Jellby:
This is at least an initial solution for #111. When a MR is created from a fork, if the `temp_branch` option is given, Marge will create a temporary branch with that name in the target project and check the CI result of this branch instead of the source project. This requires overriding the "only merge if CI passed" and "wait until CI succeeds" options of GitLab. The temporary branch is removed after a successful merge


Jellby:
Is it better like this? No changes to project settings (only to protect/unprotect the temporary branch, but the temporary branch is supposed to be disposable).

PR #115, opened 2018-07-04
Ignore GitLab server errors when looking for projects and MRs

Jellby:
It is not uncommon for GitLab to have temporary errors (500 responses or other). If this happens while processing a MR, the safe thing is probably to abort. But if it happens while looking for projects or MRs, it should be OK to just ignore the error and proceed with an empty list (i.e., wait and rerun the loop).

### DIFF
--- a/marge/merge_request.py
+++ b/marge/merge_request.py
@@ -32,10 +32,13 @@ class MergeRequest(gitlab.Resource):
 
     @classmethod
     def fetch_all_open_for_user(cls, project_id, user_id, api):
-        all_merge_request_infos = api.collect_all_pages(GET(
-            '/projects/{project_id}/merge_requests'.format(project_id=project_id),
-            {'state': 'opened', 'order_by': 'created_at', 'sort': 'asc'},
-        ))
+        try:
+            all_merge_request_infos = api.collect_all_pages(GET(
+                '/projects/{project_id}/merge_requests'.format(project_id=project_id),
+                {'state': 'opened', 'order_by': 'created_at', 'sort': 'asc'},
+            ))
+        except gitlab.InternalServerError:
+            all_merge_request_infos = []
         my_merge_request_infos = [
             mri for mri in all_merge_request_infos
             if (mri['assignee'] or {}).get('id') == user_id
--- a/marge/merge_request.py
+++ b/marge/merge_request.py
@@ -1,3 +1,4 @@
+import logging as log
 from . import gitlab
 from .approvals import Approvals
 
@@ -38,6 +39,7 @@ class MergeRequest(gitlab.Resource):
                 {'state': 'opened', 'order_by': 'created_at', 'sort': 'asc'},
             ))
         except gitlab.InternalServerError:
+            log.warning('Internal server error from GitLab! Ignoring...')
             all_merge_request_infos = []
         my_merge_request_infos = [
             mri for mri in all_merge_request_infos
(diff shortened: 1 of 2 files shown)


marge/merge_request.py
```diff
@@ -32,10 +32,13 @@ def fetch_by_iid(cls, project_id, merge_request_iid, api):
...
-        all_merge_request_infos = api.collect_all_pages(GET(
-            '/projects/{project_id}/merge_requests'.format(project_id=project_id),
-            {'state': 'opened', 'order_by': 'created_at', 'sort': 'asc'},
-        ))
+        try:
+            all_merge_request_infos = api.collect_all_pages(GET(
+                '/projects/{project_id}/merge_requests'.format(project_id=project_id),
+                {'state': 'opened', 'order_by': 'created_at', 'sort': 'asc'},
+            ))
+        except gitlab.InternalServerError:
```
JaimeLennox: Could you log a warning when this happens? I think it would be good to know about this, otherwise this could happen constantly but from Marge's point of view nothing would be wrong.

aschmolck:
What's the problem you're observing with the current behavior?

Jellby:
I had marge running on gitlab.com. There were no merge requests to process, so she was just waiting. During the night there was some problem with GitLab, marge received an internal server error and crashed. In the morning I wondered why marge wasn't running since she was just waiting. As I said above, temporary server/network problems are not so rare, and they shouldn't affect marge if she's idle.

aschmolck:
Well, there are basically two ways of  solving this:

1. making marge more robust to intermittent gitlab/network failures (but in this case, you have to add some transparent retry logic to the whole gitlab API wrapper, adding some ad hoc check to one place won't make things substantially more robust).

2. Have some external service restart marge on failure. Since marge is basically stateless[*], I'd argue this is the way to go, because it's both simpler and more general (what if your system has some other hickup, like OOM and marge is stopped for that reason?) and you can leverage all sorts of existing (logging/monitoring/specification etc.) infrastructure.

How are you running marge-bot at the moment? If you are using docker, I would suggest you specify `--restart=on-failure` (or `--restart=always`). That's basically what we're doing (and not just for marge, for essentially every important internal or external service).

If you want to run without docker, I would suggest using an init system, e.g. a systemd unit file https://patrakov.blogspot.com/2011/01/writing-systemd-service-files.html

and add
...
[Service]
Restart=on-abort   # or `always`
```
Finally, the completely minimal solution is just doing something like this:
```
while ! ./marge $MARGE_OPTIONS; do echo >&2 "Marge aborted, sleeping and restarting...." sleep 5s; done
```
You'd probably want to run that in screen, tmux or with nohup.


[*] The main downside is potentially having to reclone large repos, but in practice gitlab should hopefully not be so flaky that the overhead of this is a significant annoyance. A better fix in that case might be to make marge look for existing local clones first.

Jellby:
I'm not running with docker. Maybe I will in the future, but for testing I run it with a bare command. Sure, I don't claim my solution will solve everything, but it should fix what seems to me the most likely harmless scenario. If there is an error while marge is doing something else, I'd probably want to know, but if there's a server problem while marge is waiting, I can very well ignore it (if there's something really wrong, I'll probably notice when I try to fetch/push/browse GitLab).

aschmolck:
I'm not opposed to making marge robust to intermittent gitlab API failures, but doing so for one particular call ad hoc isn't a good idea, IMO. I just looked and I guess it's doable to implement some generic retry logic in `gitlab.py` in about 10-20 lines of code, so I might just do it. I'll need to check if there's some potential trickiness with non-idempotent actions or the few places where we basically unavoidably race because the gitlab API offers no way to do what we need in a non-racy fashion and retries would increase the window).

PR #116, opened 2018-07-18
Make --batch and --add-tested non-experimental.

aschmolck:
Achieves two things:
- can be used together (experimental was an exclusive option group)
- recognizes the fact that --batch is quite well tested now


JaimeLennox:
LGTM, aside from the tests that need fixing.

PR #121, opened 2018-08-04
adds --insecure-gitlab option

netjunki:
I'm working with a gitlab host which doesn't have a verified cert and this was a quick fix to support it.

### DIFF
--- a/marge/app.py
+++ b/marge/app.py
@@ -76,6 +76,11 @@ def _parse_config(args):
         metavar='URL',
         help='Your GitLab instance, e.g. "https://gitlab.example.com".\n',
     )
+    parser.add_argument(
+        '--insecure-gitlab',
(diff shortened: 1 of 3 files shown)


aschmolck:
I don't think this is a good idea. You should be able to use a self-signed certificate easily enough, without any code changes by doing something like this:

```
docker run -e  -v /etc/my-certs:/certs -e REQUESTS_CA_BUNDLE=/certs/my-self-signed-certificate.pem ... smarkets/marge-bot ...
```

Wouldn't that work for you?

JaimeLennox:
@aschmolck I'm not sure it's simple for everyone to setup their own certificate vs simply having an option for Marge to handle this; I'm in favour of having this given it's also behind an option, and so would have to be explicitly enabled.

PR #130, opened 2018-09-07
Job invocation isolated in a method

leguminator:
In order to enable bot derived class to launch different kind of jobs.
For example I createsd a derived Bot class that manages Rebase jobs : that new bot can overload the _get_job method in order to launch the proper job (Rebase)

### DIFF
--- a/marge/bot.py
+++ b/marge/bot.py
@@ -153,7 +153,14 @@ class Bot(object):
                 log.exception('BatchMergeJob failed: %s', err)
         log.info('Attempting to merge the oldest MR...')
         merge_request = merge_requests[0]
-        merge_job = single_merge_job.SingleMergeJob(
+        merge_job = self._get_job(
+            project=project, merge_request=merge_request, repo=repo,
+            options=self._config.merge_opts,
+        )
+        merge_job.execute()
+
+    def _get_job(self, project, merge_request, repo, options):
+        return single_merge_job.SingleMergeJob(
             api=self._api,
             user=self.user,
             project=project,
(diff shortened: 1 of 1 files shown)


marge/bot.py
```diff
@@ -153,15 +153,21 @@ def _process_merge_requests(self, repo_manager, project, merge_requests):
...
+        merge_job.execute()
+
+    def _get_job(self, project, merge_request, repo, options):
+        return single_merge_job.SingleMergeJob(
             api=self._api,
             user=self.user,
             project=project,
             merge_request=merge_request,
             repo=repo,
             options=self._config.merge_opts,
```
JaimeLennox: This should probably be `options` given you're passing it to this method; otherwise there's no point passing it down if you don't need it to be different.

marge/bot.py
```diff
@@ -153,15 +153,21 @@ def _process_merge_requests(self, repo_manager, project, merge_requests):
...
         log.info('Attempting to merge the oldest MR...')
         merge_request = merge_requests[0]
-        merge_job = single_merge_job.SingleMergeJob(
+        merge_job = self._get_job(
+            project=project, merge_request=merge_request, repo=repo,
+            options=self._config.merge_opts,
+        )
+        merge_job.execute()
+
+    def _get_job(self, project, merge_request, repo, options):
```
JaimeLennox: There's also a `BatchMergeJob`, so perhaps this method could be clearer about what type of job it's referring to? E.g. `_get_single_merge_job`.

leguminator: maybe `_get_single_job` would fit ?

marge/bot.py
leguminator: You are totally right : options=options

Issue #134, opened 2018-09-28
get_mr_ci_status() should look for any pipeline with matching SHA, not just first one


mrj10:
If I have a batch MR with multiple commits in it, my CI system may kick off pipelines for each commit, and the pipelines API call may return those pipelines in any order when get_mr_ci_status() (in ) calls Pipeline.pipelines_by_branch().

If the pipeline for the SHA we're looking for comes second or third in the list, this code won't recognize it and we'll spin forever:

```
current_pipeline = next(iter(pipelines), None)
if current_pipeline and current_pipeline.sha == commit_sha:
    ci_status = current_pipeline.status
else:
    log.warning('No pipeline listed for %s on branch %s', commit_sha, merge_request.source_branch)
    ci_status = None
```

Can we change it to this? :
```
        current_pipeline = next(iter(p for p in pipelines if p.sha == commit_sha), None)
            
        if current_pipeline is not None:
            ci_status = current_pipeline.status
        else:   
            log.warning('No pipeline listed for sha %s on branch %s; pipeline shas for branch are %s', commit_sha, merge_request.source_branch, str([p.sha for p in pipelines]))
            ci_status = None
```

JaimeLennox:
Sure, it seems reasonable to make this more explicit.

PR #135, opened 2018-09-29
Fetch pipeline with SHA matching commit SHA

JaimeLennox:
We currently just fetch the first pipeline. Whilst this is ordered by
ID, technically this doesn't guarantee that the pipeline will be for the
SHA we're looking for. To handle this, we can explicitly fetch the
pipeline for the SHA we want.

Closes: #134.

### DIFF
--- a/marge/job.py
+++ b/marge/job.py
@@ -121,9 +121,9 @@ class MergeJob(object):
             merge_request.source_branch,
             self._api,
         )
-        current_pipeline = next(iter(pipelines), None)
+        current_pipeline = next(iter(pipeline for pipeline in pipelines if pipeline.sha == commit_sha), None)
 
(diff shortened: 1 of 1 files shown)


JaimeLennox:
We always check for a particular SHA - this just makes the check less likely to fail, as we filter on the pipeline rather than always fetch the first one then compare later.

JaimeLennox:
I guess that depends on the order that GitLab returns the pipelines. I believe we currently fetch it by ID order only, which doesn't really give us any guarantees.

JaimeLennox:
And we can't always fail immediately, as sometimes a pipeline might not appear for a few seconds (so the current top pipeline is for an old commit, and we're just waiting for GitLab to create the new one).

Issue #139, opened 2018-10-09
Feature request: Docker image of Master 


grebois:
Latest tag is 3 months old, https://hub.docker.com/r/smarkets/marge-bot/tags/

There has been a lot of progress since 0.7.0, would be nice to have a tag master.

Best!

JaimeLennox:
I can look into adding this - it seems reasonable given we've taken a while to release a new version. Having said that, we'll probably release a new version quite soon!

grebois:
Awesome, thanks @JaimeLennox

PR #142, opened 2018-10-13
Push a master-tagged image

JaimeLennox:
Push a new docker image with tag "master" for every new build on the
master branch. This gives users wishing to run with the latest
development changes an easy way to stay up to date.

Note: when a tag is created, Travis sets the $TRAVIS_BRANCH envvar to
the tag, rather than the branch, hence why we only use this variable
now.

Closes: #139.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,11 @@ services:
 language: nix
 deploy:
   provider: script
-  script: make dockerize && make docker-push
+  script: make dockerize && make docker-push VERSION="$TRAVIS_BRANCH"
   on:
     repo: smarkets/marge-bot
-    tags: true
-    condition: "$TRAVIS_TAG = $(cat version)"
+    all_branches: true
+    condition: "$TRAVIS_BRANCH = $(cat version) || $TRAVIS_BRANCH = master"
 env:
   global:
     # smarkets ci docker username
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+VERSION?=$$(git rev-parse --abbrev-ref HEAD)
+
 requirements_frozen.txt requirements.nix requirements_override.nix: requirements.txt
 	pypi2nix -V 3.6 -r $^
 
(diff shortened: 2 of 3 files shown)


aschmolck:
Good idea! But now just doing make w/o passing VERSION expliciltly will misbehave. So I think something like this might be better:

At the top of the Makefil do:
```
VERSION?=$$(cat version))
```

Then only override for the master build, and leave the existing logic for proper release builds the same.

.travis.yml
```diff
@@ -4,11 +4,11 @@ services:
...
 deploy:
   provider: script
-  script: make dockerize && make docker-push
+  script: make dockerize && make docker-push VERSION="$TRAVIS_BRANCH"
   on:
     repo: smarkets/marge-bot
-    tags: true
-    condition: "$TRAVIS_TAG = $(cat version)"
+    all_branches: true
+    condition: "$TRAVIS_BRANCH = $(cat version) || $TRAVIS_BRANCH = master"
```
aschmolck: don't you want `TRAVIS_TAG` here?

JaimeLennox:
I added a stable tag, which corresponds to the latest versioned tag, and updated the README with a bit of info about this. What do you think?

Makefile
```diff
@@ -25,10 +27,15 @@ docker-push:
 	  docker login; \
 	fi
 	docker tag smarkets/marge-bot:$$(cat version) smarkets/marge-bot:latest
```
aschmolck: I think `:latest` should be `:stable` (meaning we can probably drop `:stable`, although I'm not opposed to it either). If you `docker run smarkets/marge-bot`, you should get a release version, IMO.

aschmolck:
I think the connotations of master differ by community; e.g. go's package mangement will install master, so that better be stable not bleeding edge. I'm fine w/ using it though, if it turns out to cause confusion to some users, we can change to `unstable` or whatever. I don't think `:latest` should  equal `:master` though; I think `:latest` always ought to be the latest release version, if you want bleeding edge you should have to say so explicitly, rather than the other way around where you need to say you wan `:stable`.

JaimeLennox:
Ok, I'm happy for it to behave that way too. I've left the stable tag in as it's more explicit.

Issue #159, opened 2019-01-19
Marge breaks in the presence of forks


bgamari:
The were hoping to use Marge in the [GHC project](https://gitlab.haskell.org/) to implement batched merges until proper support appears in GitLab. Unfortunately, Marge expects to be able to push to the source branch which is not easily possible in a public project like GHC where merge requests often originate from forks. It appears that the only way around this is currently to manually add the Marge bot account as a maintainer for all users' forks. Naturally, this seems quite intrusive.

I believe this would be easily avoided if Marge used GitLab's native rebase API (#157).

jcpetruzza:
The problem with using GitLab's api for rebasing is, I think, that you then can't have Marge adding trailers to the commits ("Tested-by", "Part-of", etc). But if that is a feature you weren't planning to use, then something like a `--rebase-on-gitlab` flag could probably be added which (hopefully) shouldn't be a terribly intrusive change:  Marge would still work on the local repo, and do a rebase there, but at the point of pushing the changes back to the merge request branch, she would instead request a rebase, wait for the result, and if the operation succeeds, she needs to check that the new top-commit of the MR matches the one she has on her local branch (i.e., the one that she would have otherwise pushed).

Do you think something like that would work in your setting?

bgamari:
Indeed that is what I did (albeit in a terrible, awful, no-good hacky way). See https://github.com/bgamari/marge-bot/commit/11a908b24d0dabacfab3816483d58c577654ffeb. I may try to clean this up to submit upstream once things have stabilized on our end.

Thanks for your help!

jcpetruzza:
@bgamari I gave this a shot in the PR above; let me know if works for you?

JaimeLennox:
I think this can be closed now. There is support for rebasing via the GitLab API.

Issue #157, opened 2018-12-07
Gitlab now supports rebase MR API


left4taco:
Currently the way marge-bot do rebase is to pull the branch, rebase locally on to target branch, then push it back with `--force`. But this will clear the approvals status of MR. Impersonating the original approvers is a solution but that requires admin privilege which is not always possible. 

From version 11.6, Gitlab API supports the rebase API. This is the same operation like you click the rebase button in the MR page, which makes the rebasing and reapproving a lot easier, and no more local git repo is needed.

jcpetruzza:
@left4taco I implemented this in #160, if you want to give it a try?

left4taco:
@jcpetruzza Sorry I'm too busy these days and probably cannot help😢

JaimeLennox:
Closing as this has been implemented.

PR #160, opened 2019-01-27
Support rebasing through GitLab's api

jcpetruzza:
Here's an attempt to add support for GitLab's new "merge_request/rebase" endpoint (#157), which should enable using marge in public gitlab instances (#159). It is enabled with `--rebase-remotely` and conflicts with `--use-merge-strategy` and all the `add-xxx` options.

It works by performing the rebase locally but replacing the `git push --force` step by a call to [this endpoint](https://docs.gitlab.com/ee/api/merge_requests.html#rebase-a-merge-request).

Because the change happens inside `MergeJob.update_from_target_branch_and_push()`, it wouldn't have been picked  by the mocklab tests, where this method was being replaced by a mock. So I went down a bit of a rabbit hole at first and started by adding a mock of git, to fix this. The mock works by hooking into the `Repo.git()` method and interpreting the git commands that marge emits, using an extremely simplified model of git.

Because now these tests can see inside `update_from_target_branch_and_push()`, I make it so that they run on (almost) all combinations of fusion-strategies and trailer-adding, which caught a small bug in the handling of protected branches. 

While the PR is a bit large, the individual commits should be smallish and self-contained..

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ ENV/
 .cache
 *.egg-info
 .coverage
+.pytest_cache
 
 # nix stuff
 result
--- a/marge/app.py
+++ b/marge/app.py
@@ -250,6 +250,11 @@ def main(args=None):
         if options.batch:
             logging.warning('Experimental batch mode enabled')
 
+        if options.use_merge_strategy:
+            fusion = bot.Fusion.merge
+        else:
+            fusion = bot.Fusion.rebase
+
         config = bot.BotConfig(
             user=user,
             ssh_key_file=ssh_key_file,
@@ -266,7 +271,7 @@ def main(args=None):
                 approval_timeout=options.approval_reset_timeout,
                 embargo=options.embargo,
                 ci_timeout=options.ci_timeout,
-                use_merge_strategy=options.use_merge_strategy,
+                fusion=fusion,
             ),
             batch=options.batch,
         )
--- a/marge/app.py
+++ b/marge/app.py
@@ -118,7 +118,7 @@ def _parse_config(args):
             "built-in rebase functionality, via their API. Note that Marge can't add\n"
             "information in the commits in this case.\n"
         ),
-    ),
+    )
     parser.add_argument(
         '--add-tested',
         action='store_true',
--- a/marge/app.py
+++ b/marge/app.py
@@ -226,7 +226,7 @@ def _secret_auth_token_and_ssh_key(options):
 
 
 def main(args=None):
-    if not args:
+    if args is None:
         args = sys.argv[1:]
     logging.basicConfig()
 
--- a/marge/app.py
+++ b/marge/app.py
@@ -111,6 +111,15 @@ def _parse_config(args):
         ),
     )
     parser.add_argument(
+        '--rebase-remotely',
+        action='store_true',
+        help=(
+            "Instead of rebasing in a local clone of the repository, use GitLab's\n"
+            "built-in rebase functionality, via their API. Note that Marge can't add\n"
+            "information in the commits in this case.\n"
+        ),
+    ),
+    parser.add_argument(
         '--add-tested',
         action='store_true',
         help='Add "Tested: marge-bot <$MR_URL>" for the final commit on branch after it passed CI.\n',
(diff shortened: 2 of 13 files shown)


mpickering:
This won't fix the batch job working with forks. So won't fix the real problem which is present in #157

jcpetruzza:
@mpickering IIUC, in batch mode, after a batch passes the tests, it is first pushed to master on the *target* branch and then all the PRs are rebased, and because the changes are already there, gitlab notices the PR is a no-op and closes it. If the rebasing of the PR is done through the API I expect the same to happen.

I'm probably missing something?

mpickering:
Yes sorry, the problem is that after rebasing `_repo.fast_forward` tries to fetch the wrong branch as `local` isn't passed to it.

My attempted fix is something like - https://github.com/mpickering/marge-bot/commit/c5b0c598849648d25fee1ca6d356ef10f7d39172

jcpetruzza:
I see what you mean now! Yes, the batch processing code seems to be assuming that the rebased branch will be already in the repository (which made sense when forks where not handled) and, when using gitlab-rebase, we'd need to fetch the changes beforehand as you did there.

@mpickering do you want to add  a commit on top of this PR?

mpickering:
Does my change look correct? It felt a bit hacky to me. I'm also unsure how to test this locally, I just deployed my fix and have yet to see if it actually works.

mpickering:
@jcpetruzza Can you please tell me how to run the tests so that I can add a test where a MR comes from a fork? It has been a nightmare trying to get this to work with a 6hr CI turnaround.

mpickering:
I tried just running `nix-build` as CI does but from a fresh checkout it fails nearly immediately for me 🤷‍♂️

jcpetruzza:
@mpickering You can do `nix-shell --run py.test` to run everything. When working on this, I would launch a `nix-shell` and  run `py.test tests/test_foo.py`, and for a faster feedback loop, you can restrict the tests to run with `-k test_name`. You may want to edit `setup.py` and temporarily comment out the line that says:
```
addopts = --flake8 --pylint --cov=marge
```
since pylint is super slow (there must be a better way of doing this, but well)

jcpetruzza:
Also, instead of doing the checkout yourself as in your commit, would it work if, whenever the rebase was done by gitlab, you pass the `source_repo_url`  to `fast_forward()` method (and don't use `local=True`)? If I'm reading the code correctly, this would first do the `git fetch` as is needed in this case.

mpickering:
@jcpetruzza I will try and add some tests tonight and once I have added those, will feel a lot more confident refactoring.

mpickering:
Just to update this. I tried adding some tests but it was too complicated for me and the tests for batch mode are quite lacking to base things from. 

However, I eventually got things to work properly but you should be aware that there is a race condition present in the rebase API where the API will say that the rebase has concluded but the change is not reflected yet in the remote. I had to add a big sleep (2 minutes) to be sure that this situation didn't happen.

You can see the changes I made on my branch if you're interested in merging them into upstream. I'm very reluctant to do any refactoring now though due to the lack of tests.

https://github.com/mpickering/marge-bot

kgilden:
@mpickering @jcpetruzza I'm trying it out and indeed it seems to hit a race condition. It basically tries to rebase twice, which in our case causes the pipeline to be skipped. That in turns makes it possible to immediately merge the branch before waiting until the tests on the branch actually pass. Please see the screenshot below.

2019-04-11-110340_286x322_scrot

Any ways to work around this?

mpickering:
We are using a fork of marge now (in order to fix the batch merge mode). The only workaround I know is to increase the timeouts.

Okhoshi:
The problem of race condition is due to Gitlab not returning `rebase_in_progress` on Get Single MR by default. As it is an asynchronous operation, you need to provide `include_rebase_in_progress` parameter when querying the MR state (as per https://docs.gitlab.com/ee/api/merge_requests.html#rebase-a-merge-request).

kdsnice:
#348 adding `include_rebase_in_progress` parameter and prevents marge-bot to complain about someone skipping the queue on each merge request

Issue #156, opened 2018-12-07
Membership in groups not found.


Johannestegner:
Marge bot seems to fail to check for MR's in nested projects where the user have group access but not specifically to the project in question.

Checking the gitlab API we can see that even if the User does indeed have access (in tested case "owner") the resulting JSON payload won't have its permissions correctly set.

```
"permissions":{"project_access":null,"group_access":null}
```

This is in essence a issue with gitlab, as one would expect the API to actually give the correct information and a bug report on the gitlab issue tracker exists: https://gitlab.com/gitlab-org/gitlab-ce/issues/46510

I would propose a temporary solution like a flag which ignores checking permission until the issue is fixed on gitlabs end. This is of course a bit unsafe could possibly break the bot while it's working, but naming the flag something like `--UNSAFE-skip-permission-check` and print some warnings on startup should at the least make this obvious to users.

PR #166, opened 2019-03-09
 Only fetch projects with min access level

JaimeLennox:
If we're on an appropriate GitLab version, that is. GitLab doesn't return the right permissions for projects belonging to the user if the project is in a subgroup, and Marge is set as a member of the top-level group[1]. This is rather annoying as Marge will then complain that she doesn't have permissions for a project when she does.

A new field available in GitLab version >= 11.2 allows us to only fetch projects for which we have a certain access level. This returns the right projects even with nested groups, so we can prefer this method if it is available.

Fixes: #156

### DIFF
--- a/marge/bot.py
+++ b/marge/bot.py
@@ -91,7 +91,7 @@ class Bot(object):
         for project in projects:
             project_name = project.path_with_namespace
 
-            if project.access_level.value < AccessLevel.reporter.value:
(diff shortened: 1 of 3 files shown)


aschmolck:
Very annoying they still haven't fixed this issue upstream, 9 months later, so although I'm not too keen to accumulate more and more workarounds for gitlab bugs let's merge this.

Issue #167, opened 2019-03-15
Application crash: GitLab failed to provide user permissions on project


Johannestegner:
**What happens**: In latest master build, the application crash when fetching permissions for a project. The assertion is probably quite correct in essence, but maybe it should not actually crash the application all together?

**What did I expect**: Possibly a error message or info that the "user" have no access, not a full application crash.

**Version**: `master`

---

The following line is the assertion call:

https://github.com/smarkets/marge-bot/blob/2b7531e8bb655ac3a7c42c71de22f81486ceb6ca/marge/project.py#L88

Full callstack:

```
2019-03-15 13:19:49,918 INFO Finding out my current projects...
Traceback (most recent call last):
  File "/nix/store/clzhpjjcxsl521f96lcrgfcgn1yxvahs-python3.6-marge-0.8.1/bin/.marge.app-wrapped", line 4, in <module>
    marge.app.main()
  File "/nix/store/clzhpjjcxsl521f96lcrgfcgn1yxvahs-python3.6-marge-0.8.1/lib/python3.6/site-packages/marge/app.py", line 307, in main
    marge_bot.start()
  File "/nix/store/clzhpjjcxsl521f96lcrgfcgn1yxvahs-python3.6-marge-0.8.1/lib/python3.6/site-packages/marge/bot.py", line 42, in start
    self._run(repo_manager)
  File "/nix/store/clzhpjjcxsl521f96lcrgfcgn1yxvahs-python3.6-marge-0.8.1/lib/python3.6/site-packages/marge/bot.py", line 60, in _run
    projects,
  File "/nix/store/clzhpjjcxsl521f96lcrgfcgn1yxvahs-python3.6-marge-0.8.1/lib/python3.6/site-packages/marge/bot.py", line 94, in _process_projects
    if project.access_level < AccessLevel.reporter:
  File "/nix/store/clzhpjjcxsl521f96lcrgfcgn1yxvahs-python3.6-marge-0.8.1/lib/python3.6/site-packages/marge/project.py", line 88, in access_level
    assert effective_access is not None, "GitLab failed to provide user permissions on project"
AssertionError: GitLab failed to provide user permissions on project
```

PR #168, opened 2019-03-16
Fix accessing permissions when using min_access_level

JaimeLennox:
We fetch projects using the `min_access_level` field to bypass a bug in
GitLab causing no permissions to be provided in some cases. When this
happens, retrieving the access level is not possible because of the same
bug meaning no permissions are available.

To workaround this, we can add a fallback permission with developer
access, since at this point we know the projects we have fetched have at
least this permission.

Fixes: https://github.com/smarkets/marge-bot/issues/167


aschmolck:
Merging (sigh). We should maybe add an issue to revert these MRs once gitlab finally fixes this upstream.

PR #175, opened 2019-04-13
Fix removing source branches after merge

glasah:
From api docs looks like wrong param

```
remove_source_branch	boolean	no	Flag indicating if a merge request should remove the source branch when merging
``

https://docs.gitlab.com/ee/api/merge_requests.html#create-mr


JaimeLennox:
Thanks!

PR #177, opened 2019-04-23
By default, ignore archived projects.

### LINKED ISSUE #172: Feature request: exclude archived projects
Currently, `marge` will look for MRs in all repos that match a regexp, whether the repo is archived or not. It would be nice to have a config switch to exclude archived projects, or just exclude them by default.

snim2:
Closes #172 

This is a first go at excluding archived projects; it likely needs better testing before it can be merged.

This PR adds a new CLI argument (`--include-archived`) and ENV VAR (`MARGE_INCLUDE_ARCHIVED`) which defaults to `False`. If `True`, Marge will process archived projects (which cannot have their MRs merged) as well as live projects (which can).

The API for `Project.fetch_all_mine()` has changed to include an optional argument `include_archived` which matches the CLI arg.

The current default behaviour is for Marge to process all projects, which wastes time and resources, linearly with the number of archived projects.

### DIFF
--- a/marge/app.py
+++ b/marge/app.py
@@ -168,6 +168,12 @@ def _parse_config(args):
         help="Only process projects that match; e.g. 'some_group/.*' or '(?!exclude/me)'.\n",
     )
     parser.add_argument(
+        '--include-archived',
+        type=bool,
+        default=False,
+        help="Marge should process all available projects, including those that are archived.\n",
+    )
+    parser.add_argument(
         '--ci-timeout',
         type=time_interval,
         default='15min',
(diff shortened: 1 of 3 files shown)


aschmolck:
Thanks for your contribution, and sorry for the delay!

My only review comment would be that option to inspect archived projects is not needed, if you remove it this is good to go in (i.e. archived projects should always be excluded, I don't see a case where overriding this would be useful).

snim2:
Many thanks, I've pushed a commit to make that change. Do you want me to squash?

aschmolck:
Yes, please!

aschmolck:
Once it's squashed, I'll merge.

aschmolck:
Great – thanks!

PR #178, opened 2019-05-03
nix: formatting, smarter `src` in marge.nix

stolyaroleh:
`src = ./.;` would include everything, including the autogenerated `result` symlink and `.git`. This PR limits it to a required subset of files; changing README or reformatting `*.nix` won't trigger a rebuild anymore.


JaimeLennox:
Added as part of #179

PR #183, opened 2019-05-07
nix: delete pinnedNixpkgs.nix

stolyaroleh:
`nixpkgs` is now pinned in `nix/sources.json`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ ENV/
 
 # nix stuff
 result
+result-*
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ marge-bot:
 
 .PHONY: clean
 clean:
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -6,4 +6,4 @@ ENV/
 .pytest_cache
 
 # nix stuff
-result
```
aschmolck: different glob to rm -rf one I think we need two lines (`result` `result-*`)?

Makefile
```diff
@@ -9,7 +9,7 @@ marge-bot:
 
 .PHONY: clean
 clean:
-	rm -rf .cache result requirements_frozen.txt
+	rm -rf .cache result-* requirements_frozen.txt
```
aschmolck: should be `... result result-* ...` right?

PR #186, opened 2019-05-11
Update to support multiple assignees

benjamb:


aschmolck:
Thanks, looks good to me!

PR #191, opened 2019-05-15
Update API for remove source branch on merge.

Viatorus:
To accept an MR and remove the source branch one has to set `should_remove_source_branch` like mention here:
https://docs.gitlab.com/ee/api/merge_requests.html#accept-mr

I also checked older versions (10.3). It was always `should`. Maybe in v3 it was different.


JaimeLennox:
Ah, my bad. I accepted #175 with a quick look at the docs, but it seems that was for creating an MR, not accepting, so I believe you are right. It's quite annoying that this parameter changes name between these two calls.

Issue #202, opened 2019-06-11
flood when CI failed


micheelengronne:
Marge-bot sometimes floods comments of a MR with "I couldn't merge this branch: CI failed!" when the CI failed on a MR with marge-bot assigned.

It generally happens when trying to merge a MR created by [renovate](https://github.com/renovatebot/renovate) and using the same bot account.

Here is the log:
``
2019-06-11 08:54:58,436 INFO Processing !43 - 'feat: Update **** Docker tag to v1.35.0'
2019-06-11 08:54:58,921 INFO Ensuring MR !43 is mergeable
2019-06-11 08:54:58,948 INFO Running git -C /tmpxxo1dnyv/tmpupfoojcg fetch --prune origin
2019-06-11 08:54:59,276 INFO Running git -C /tmpxxo1dnyv/tmpupfoojcg checkout -B renovate/**** origin/renovate/**** --
2019-06-11 08:54:59,399 INFO Running git -C /tmpxxo1dnyv/tmpupfoojcg rebase origin/master
2019-06-11 08:54:59,474 INFO Running git -C /tmpxxo1dnyv/tmpupfoojcg rev-parse HEAD
2019-06-11 08:54:59,479 INFO Running git -C /tmpxxo1dnyv/tmpupfoojcg rev-parse origin/master
2019-06-11 08:54:59,484 INFO Adding trailers for MR !43
2019-06-11 08:54:59,484 INFO Running git -C /tmpxxo1dnyv/tmpupfoojcg filter-branch --force --msg-filter 'TRAILERS='"'"'Tested-by: BOT <https://****/merge_requests/43>'"'"' python3 /nix/store/rsc74jcy093n1qdiqda61sj3valz7839-python3.6-marge-0.8.1/lib/python3.6/site-packages/marge/trailerfilter.py' 'renovate/****^..renovate/****'
2019-06-11 08:54:59,632 INFO Running git -C /tmpxxo1dnyv/tmpupfoojcg rev-parse HEAD
2019-06-11 08:54:59,637 INFO Running git -C /tmpxxo1dnyv/tmpupfoojcg filter-branch --force --msg-filter 'TRAILERS='"'"'Part-of: <https://****/merge_requests/43>'"'"' python3 /nix/store/rsc74jcy093n1qdiqda61sj3valz7839-python3.6-marge-0.8.1/lib/python3.6/site-packages/marge/trailerfilter.py' origin/master..renovate/****
[...]

Issue #189, opened 2019-05-14
Maximum number of reassigns 


sobolevn:
When `marge` user creates a merge request and is assigned to merge it, infinite loop happens.

<img width="923" alt="Снимок экрана 2019-05-14 в 16 00 05" src="https://user-images.githubusercontent.com/4660275/57699856-66cb1000-7661-11e9-8e64-96ec902cecb3.png">
<img width="1188" alt="Снимок экрана 2019-05-14 в 15 59 46" src="https://user-images.githubusercontent.com/4660275/57699857-66cb1000-7661-11e9-94f9-b137eb9b50d6.png">

And there's no stopping mechanism. I suggest to add something like `--max-reassigns` option to stop it.

Or to check if future assignee and author is the same user. It might be easier.

aschmolck:
I think your suggestion of marge-bot not unassigning itself just to re-assigning to itself is the right one, I'd be happy to accept a PR to that effect.

This is just a minor issue, because it shouldn't happen outside people testing their marge-bot setup, but I agree it would nonetheless good to fix.

micheelengronne:
Related https://github.com/smarkets/marge-bot/issues/202

micheelengronne:
MR https://github.com/smarkets/marge-bot/pull/203

PR #203, opened 2019-06-11
not unassigning itself if MR author

micheelengronne:
Marge-bot will not unassign/reassign itself if it is the author.

Trying to solve https://github.com/smarkets/marge-bot/issues/189 and https://github.com/smarkets/marge-bot/issues/202

### DIFF
--- a/marge/merge_request.py
+++ b/marge/merge_request.py
@@ -171,7 +171,7 @@ class MergeRequest(gitlab.Resource):
         ))
 
     def assign_to(self, user_id):
-        if self.author_id() not in self.assignee_ids():
+        if self.author_id not in self.assignee_ids:
             return self._api.call(PUT(
                 '/projects/{0.project_id}/merge_requests/{0.iid}'.format(self),
                 {'assignee_id': user_id},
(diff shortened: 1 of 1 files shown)


aschmolck:
Looking at the code, it doesn't seem to me that this changes existing behavior, since the only place these functions should be called for is unassign_from_mr which looks as follows:

```
marge/job.py-    def unassign_from_mr(self, merge_request):
marge/job.py-        log.info('Unassigning from MR !%s', merge_request.iid)
marge/job.py-        author_id = merge_request.author_id
marge/job.py-        if author_id != self._user.id:
marge/job.py:            merge_request.assign_to(author_id)
marge/job.py-        else:
marge/job.py-            merge_request.unassign()
```

This also means I don't understand why the reassign spam is actually happening in the first place. Did I overlook something?

micheelengronne:
The current patch does not seem to work. I am investigating.

micheelengronne:
The problem is not assign/unassign but the fact that marge-bot does not check if a comment was already made for this build and this status. It should check that instead.

micheelengronne:
I will write another PR.

Issue #205, opened 2019-06-12
branch-regexp but for source branch


micheelengronne:
Is it possible to add an option, source-branch-regexp that works like branch-regexp but for source branches ?

aschmolck:
I'm happy to accept a MR that does this, since I think it might have other uses than running all bots under a single account (which is still not something I feel to be very desirable). Marge-bot should probably leave a single comment on the MR that it will be ignored because it doesn't match source-branch-regexp to prevent potential confusion.

micheelengronne:
MR https://github.com/smarkets/marge-bot/pull/206

PR #206, opened 2019-06-12
source branch regexp added

micheelengronne:
Source branch regexp option added.

Answer https://github.com/smarkets/marge-bot/issues/205

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -216,6 +216,12 @@ def test_branch_regexp():
             assert bot.config.branch_regexp == re.compile('foo.*bar')
 
 
+def test_source_branch_regexp():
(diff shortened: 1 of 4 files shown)


tests/test_app.py
```diff
@@ -216,6 +216,12 @@ def test_branch_regexp():
             assert bot.config.branch_regexp == re.compile('foo.*bar')
 
 
+def test_source_branch_regexp():
```
aschmolck: I think what should be tested is that by default all source branches are processed, but if a filter is explicitly specified only source branches that match is get merged.

micheelengronne:
It was locally tested and it works.

aschmolck:
Cool, merged -- thanks!

PR #210, opened 2019-06-15
antiflood option for comments

micheelengronne:
Attempt to solve https://github.com/smarkets/marge-bot/issues/189

Add an option `--comment-antiflood` to avoid a flood with the same comment.

### DIFF
--- a/marge/app.py
+++ b/marge/app.py
@@ -174,6 +174,12 @@ def _parse_config(args):
         help='How long to wait for CI to pass.\n',
     )
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


micheelengronne:
This patch checks that the last comment body is not the same as the one that would be written.

It works.

The only anti-pattern that I can see is when an event is triggered that does not add any note to the MR (like retriggering manually the pipeline).

micheelengronne:
Hum, I have to pass more time figuring how the mock works to add the new endpoint. Any hint ?

PR #211, opened 2019-06-15
Fix unassign when author is marge

JaimeLennox:
According to the docs:

"Set to 0 or provide an empty value to unassign all assignees."[1]

So setting to `None` seems like it would work, although in practice it doesn't.

I've manually tested with `0` and it works fine.

[1] https://docs.gitlab.com/ee/api/merge_requests.html#update-mr

Fixes: #189

### DIFF
--- a/marge/merge_request.py
+++ b/marge/merge_request.py
@@ -177,7 +177,7 @@ class MergeRequest(gitlab.Resource):
         ))
 
     def unassign(self):
-        return self.assign_to(None)
+        return self.assign_to(0)
 
     def fetch_approvals(self):
         # 'id' needed for for GitLab 9.2.2 hack (see Approvals.refetch_info())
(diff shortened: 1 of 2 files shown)


micheelengronne:
The problem is that it remove the PR from any assignee. Making it lost in the void :)

micheelengronne:
I am working on a solution to avoid the flood that keeps assignees https://github.com/smarkets/marge-bot/pull/210 I just have to figure out how to add an endpoint in the mock.

aschmolck:
@micheelengronne Removing the PR from any assignee (and maybe leaving a comment about user error) is the correct behaviour.  Marge-bot is not meant to be the author of PRs, and although this should fail gracefully (maybe with an explanatory note), I'm not interested in code support beyond this, since it's nothing that can ever work particularly well in the first place.

Having said that, if there is something that would be desirable for other reasons and helps this unsupported use case I might still consider merging it.

Issue #185, opened 2019-05-10
GitLab pipeline API change: detached pipeline refs changed - Marge fails to detect pipeline


veeg:
Due to recent [GitLab API Change](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/25504), the refs slug received from [list pipelines](https://docs.gitlab.com/ee/api/pipelines.html#list-project-pipelines) returns a ref changed to format `refs/merge-requests/[MERGE_REQUEST_NUMBER]/head`, instead of `[BRANCH_NAME]`.

This is only applicable when the pipeline is ran on merge_requests, e.g., detached.

The following call needs to be augmented with a refs argument:
https://github.com/smarkets/marge-bot/blob/183be28fcac622c006dd81caa4fa62c2298d4834/marge/job.py#L138-L142

veeg:
This API change has renderes Margebot useless for our deployment.

Does anyone have a suggestion on how this scenario should be handled, and I can have a stab at implementing it.

dbinderAtCP:
We are stymied by this too. One thought could be to get all of the pipelines associated with the project_id, by adding an all_pipelines() method to Pipeline,  requiring the iterator in get_mr_ci_status() looking for the matching sha to deal with the longer list.

PR #212, opened 2019-06-16
Fetch pipelines by merge request instead of branch

JaimeLennox:
... if possible.

Since we operate on merge requests this makes more sense, but the feature was only available from version 10.5.

It allows us to fetch all pipelines for a particular merge request, including detached ones, which would otherwise not be fetched due to a different `ref` format.

An annoyance of this route is that it doesn't allow ordering or sorting, which we use to determine the latest pipeline by. However it's very simple to just do this ourselves; the only downside is this is no longer
customizable, but nothing in the codebase appears to use this feature anyway.

The batch job tries cancel pipelines by their status, which we can't filter by for this route (unless we do it manually as well) so I've left that using the old method. This can mean some detached pipelines may
continue running after a batch job has completed, but this seems like a very minor issue.

Fixes: #185

### DIFF
--- a/marge/job.py
+++ b/marge/job.py
@@ -135,11 +135,19 @@ class MergeJob:
     def get_mr_ci_status(self, merge_request, commit_sha=None):
         if commit_sha is None:
             commit_sha = merge_request.sha
-        pipelines = Pipeline.pipelines_by_branch(
-            merge_request.source_project_id,
-            merge_request.source_branch,
-            self._api,
-        )
+
+        if self._api.version().release >= (10, 5, 0):
+            pipelines = Pipeline.pipelines_by_merge_request(
+                merge_request.source_project_id,
+                merge_request.iid,
+                self._api,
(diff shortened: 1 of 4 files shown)


b1rdex:
This causes an error if ci is configured to run different tasks both on the branch and the merge request.

I have tests and all other things configured to be run on the branches because such tasks don't need a merge request to be run. Also, I have one task configured to be run on merge requests — it is used to toggle merge request settings (enable squash). So now Marge thinks that this simple task is my tests and merges even if actual tests are running and not finished yet.

Looks like that's because `current_pipeline = next(iter(pipeline for pipeline in pipelines if pipeline.sha == commit_sha), None)` line — Marge takes just one pipeline in accordance.

b1rdex:
Could we please make this behavior switchable through the config file?

b1rdex:
Extracted issue from comments to #227.

PR #213, opened 2019-06-19
Update changelog and bump version to 0.9.0

michaelmalonenz:
I believe that this should be enough to get a new docker image built?  I'm hoping to get the fix in #212 available for the docker image.

Let me know if there's anything else I need to do - like making a tag, etc.


thinkl33t:
You'll need to add a 0.9.0 tag too for the CI to run i believe

michaelmalonenz:
Thinking about it a bit more - do I even have permission to tag?  I don't have write access which makes me think I don't.  And tags can't be part of the CI.

Also, shouldn't the tag be on actual master or a release branch?

thinkl33t:
I believe the process is to tag the merge commit when this is merged - but someone with commit access would have to do it.

JaimeLennox:
Since 0.8.2 was never released I've superceded it with 0.9.0 (#216). But thanks for pushing the release - sorry it took so long!

michaelmalonenz:
No worries about the time - you don't owe us anything.  And thank you for the project, it helps us out a bunch!

PR #233, opened 2019-10-08
Add Only allow merge requests to be merged if the pipeline succeeds requirement to readme

mathyoudawson:
**Issue**
Margebot requires the Only allow merge requests to be merged if the pipeline succeeds setting to be enabled for both single branch merging and batch merging. I could not find any documentation on this. I also wonder if this method should be changed such that if master is protected this setting doesn't need to enabled as margebot would implicitly enforce this.

I have fixed a link to semi-linear history and fast-forward in the Readme as this link went to the main merge requests page rather than the specific section.

I also added a sentence in the `Running` section explaining that this setting must be enabled with a link to how this is done.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,8 @@ Once you have the token, put it in a file, e.g. `marge-bot.token`.
 
 Finally, create a new ssh key-pair, e.g like so
 
+Margebot also requires the [Only allow merge requests to be merged if the pipeline succeeds](https://docs.gitlab.com/ee/user/project/merge_requests/merge_when_pipeline_succeeds.html#only-allow-merge-requests-to-be-merged-if-the-pipeline-succeeds) setting to be enabled.
+
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -183,6 +183,8 @@ Once you have the token, put it in a file, e.g. `marge-bot.token`.
 
 Finally, create a new ssh key-pair, e.g like so
 
+Margebot also requires the [Only allow merge requests to be merged if the pipeline succeeds](https://docs.gitlab.com/ee/user/project/merge_requests/merge_when_pipeline_succeeds.html#only-allow-merge-requests-to-be-merged-if-the-pipeline-succeeds) setting to be enabled.
```
aschmolck: "To keep your master green, marge-bot also requires"

You can run marge-bot without only allow merges when pipeline succeeds in the slightly atypical case that you only care about the reviewer tagging feature (for auditing requirements, say).

Issue #164, opened 2019-03-06
When using the `--batch` feature the commit hashes that end up on master are not the same hashes that tests were run on


Dalamar42:
### Our workflow

First to explain our workflow so the issue makes sense:

Each run of our CI produces docker images that are tagged with the commit of the respective HEAD. Those images are then published to our internal docker registry and are what is shipped to production.
 
Our CI contract has three strong assumptions:

1. CI is green on every branch before that branch is merged on master
1. everything on master is immediately shippable
1. we ship exactly what we test.

### Issue

When using the `--batch` feature the commit hashes that end up on master are not the same hashes that tests were run on. The workflow currently looks like this:

1. CI runs on individual branches and they are assigned to Marge
1. Marge creates a batch branch
1. CI runs on the batch branch and passes
1. Marge rebases original MRs on master and applies trailers (reviewed-by, etc.) (not sure about the ordering here)
1. Marge merges on master

With this workflow, at the time that the branches are merged on master CI has been run against the HEADs of the original branches as well as the HEAD of the batch branch. However, that may not be the same commit hash as the HEAD of master after this process is complete.

This issue breaks the second part of our CI contract. As no CI has been run on the current HEAD of master there are no images associated with it and we need to wait for CI to run once more on master after the merge for this commit to be shippable.

To avoid this issue what we would like is for the commit hashes that end up on master to be the same as those that were tested on the batch branch.

### Proposal

1. Marge creates a batch branch
1. Apply trailers optimistically
1. Run CI on batch branch
1. If it succeeds, rewrite the original branches by cherry-picking the commits from the batch branch and merge them

In the unlikely case the trailers have changed on any one of the original branches (e.g. the list of approvers has changed) abort the batch.

PR #252, opened 2020-04-03
Optimistic batching for batch merging

deibido:
Fix for issue #164.

Previously we added trailers to each MR and
merged the constituent MRs of a batch MR to
the target branch.
Now we will now optimistically add the trailers
to each MR and rebase on the batch branch. And
when the batch branch passes CI we merge it
which will automatically merge each constituent
branch.


marge/batch_job.py
```diff
@@ -225,36 +215,34 @@ def execute(self):
                 working_merge_requests.append(merge_request)
         if len(working_merge_requests) <= 1:
             raise CannotBatch('not enough ready merge requests')
-        if self._project.only_allow_merge_if_pipeline_succeeds:
```
JaimeLennox, line 228: We seem to have lost this check. Whilst batching is a bit weird if you don't have pipelines enabled, we'd still want to support simply merging the batch branch in without waiting for any pipelines (which won't exist now) to finish.

marge/batch_job.py
```diff
@@ -207,6 +194,9 @@ def execute(self):
                     local=True,
                 )
 
+                # Ensure that individual branch commit SHA matches matches that of its equivalent in batch MR
+                self.push_force_to_mr(merge_request, True, source_repo_url, skip_ci=True)
```
JaimeLennox, line 198: I'm not sure we want to push here - the source branch for the MR will now contain commits of other MRs that have previously been added to the batch branch - we don't want to mess with the original MRs here, especially as the batch can fail.

marge/batch_job.py
```diff
@@ -225,36 +215,34 @@ def execute(self):
...
+                        batch_mr_iid=batch_mr.iid,
+                        error=err.reason,
+                    ),
+                )
+            raise CannotBatch(err.reason) from err
         for merge_request in working_merge_requests:
             try:
                 # FIXME: this should probably be part of the merge request
                 _, source_repo_url, merge_request_remote = self.fetch_source_project(merge_request)
-                self.ensure_mr_not_changed(merge_request)
```
JaimeLennox, line 251: Why has this check been lost?

marge/batch_job.py
```diff
@@ -198,6 +182,9 @@ def execute(self):
                     merge_request.source_branch,
                     '%s/%s' % (merge_request_remote, merge_request.source_branch),
                 )
+                # Apply the trailers before running the batch MR
+                self.add_trailers(merge_request)
+                self.push_force_to_mr(merge_request, True, source_repo_url, skip_ci=True)
```
JaimeLennox, line 187: It would be nicer to specify keyword args for the boolean, as makes it easier to know what this is doing (i.e. `branch_was_modified=True`)

marge/git.py
```diff
@@ -146,7 +146,14 @@ def push(self, branch, *, source_repo_url=None, force=False):
...
         force_flag = '--force' if force else ''
-        self.git('push', force_flag, source, '%s:%s' % (branch, branch))
+        # The following needs to be split into 2 variables as any whitespace in the string
+        # causes it to be quoted which makes the string ignored by git
+        if skip_ci:
+            skip_1 = '-o'
+            skip_2 = 'ci.skip'
+        else:
+            skip_1, skip_2 = '', ''
+        self.git('push', force_flag, skip_1, skip_2, source, '%s:%s' % (branch, branch))
```
JaimeLennox, line 156: I think you can just have:

 `skip_flag = ['-o', 'ci.skip'] if skip_ci else []`

and then:

`self.git('push', force_flag, *skip_flag, source, ...)`

PR #257, opened 2020-05-05
modern version of the deploy file

hiloyt:
apps/v1 was introduced in 1.9

extensions/v1beta1, apps/v1beta1 and apps/v1beta2 are deprecated (and not used anymore after 1.15)

 Time to go into the future! :)


sysadmiral:
I would suggest an accompanying update to the docs to show that the deploy file is for `kubernetes version > 1.15` :)

PR #256, opened 2020-04-26
Add argument to disable fast forwarding of batch merges

mathyoudawson:
Previously when merging batches fast forwarding is enabled. This is
problematic when reverting changes that have been fast forwarded as
you then have to revert each individual commit.

This commit adds an argument to disable the fast forward merging of
batches.

Fixes #255

PR #259, opened 2020-05-12
Pass use_no_ff_batches into merge options

mathyoudawson:
#256 introduced an error where the use use_no_ff_batches option was not passed into the
options correctly. Merge options are passed into the batch job class,
and use_no_ff_batches was not included in the merge options.

Thus creating the error:
AttributeError: 'MergeJobOptions' object has no attribute 'use_no_ff_batches'

This PR adds use_no_ff_batches to the merge options.

### DIFF
--- a/marge/app.py
+++ b/marge/app.py
@@ -311,6 +311,7 @@ def main(args=None):
                 embargo=options.embargo,
                 ci_timeout=options.ci_timeout,
                 fusion=fusion,
+                use_no_ff_batches=options.use_no_ff_batches,
             ),
             batch=options.batch,
         )
--- a/marge/job.py
+++ b/marge/job.py
@@ -408,6 +408,7 @@ JOB_OPTIONS = [
     'embargo',
     'ci_timeout',
     'fusion',
+    'use_no_ff_batches',
 ]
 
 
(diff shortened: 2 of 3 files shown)


RobertKirk:
Do you know why the tests you wrote for #256 didn't catch this error? It would be good to cover that, so we can't have a regression.

mathyoudawson:
The `merge_batch` method i introduced was tested, but the caller was not. I was surprised that other tests for the `accept_mr` method did not pick this up. Will have a go at adding some tests for this :)

RobertKirk:
Great, thanks. It's possible the caller's test mocks the `merge_batch` method.

mathyoudawson:
@RobertKirk Sorry this has taken me ages.
I've been struggling to wrap my head around how these mocks work.

Every time `accept_mr` is called during testing the exception `Someone pushed to branch while we were trying to merge` is raised.
This is due to `actual_sha` being bound to a mock object `<MagicMock name='mock.rebase()' id='4656111568'>`
https://github.com/smarkets/marge-bot/blob/master/marge/batch_job.py#L139

Python is not a language I know very well so I feel some of this is going over my head. Would appreciate some help or a push in the right direction

RobertKirk:
I think looking at it again I'm happy to just merge it in. I agree the mocks are confusing.

mathyoudawson:
@RobertKirk Cheers for merging this. For what it's worth the company I work for has been running the above code in batch mode for the last week with no issues.

PR #264, opened 2020-07-28
Improve batching: allow batch MRs use merge commits

tclh123:
Previously, we applies the commits in individual MRs to the batch MR linearly, so the commits will be rewritten and the commit sha will not be the same one passed CI, which is not good.

To make the commits be exactly the same in the individual MRs and the batch MR, I added an option to allow using merge commits in batch MRs. It acts as a helper commit in the temp branch (the batch branch) which won't affect the final merge method (we usually use the fast forwarding merge).
Besides, updating (rebase master and apply trailers) each MR before we create the batch MR to avoid rewrite commits after.

- Add `--use-merge-commit-batches` option: Use merge commit when creating batches, so that the commits in the batch MR will be the same with in individual MRs.
- Add `--skip-ci-batches` option: Skip CI when updating individual MRs when using batches. Since we have tested the batch MR, no need to run CI on individual MRs any more after we rewrite the commits.

Fixes https://github.com/smarkets/marge-bot/issues/164 https://github.com/smarkets/marge-bot/pull/252

### DIFF
--- a/marge/app.py
+++ b/marge/app.py
@@ -217,7 +217,7 @@ def _parse_config(args):
         '--use-merge-commit-batches',
         action='store_true',
         help='Use merge commit when creating batches, so that the commits in the batch MR '
-             'will be the same with in individual MRs\n',
+             'will be the same with in individual MRs. Requires sudo scope in the access token.\n',
     )
     parser.add_argument(
         '--skip-ci-batches',
--- a/marge/app.py
+++ b/marge/app.py
@@ -213,6 +213,17 @@ def _parse_config(args):
         action='store_true',
         help='Disable fast forwarding when merging MR batches'
     )
+    parser.add_argument(
+        '--use-merge-commit-batches',
+        action='store_true',
+        help='Use merge commit when creating batches, so that the commits in the batch MR '
+             'will be the same with in individual MRs\n',
+    )
+    parser.add_argument(
+        '--skip-ci-batches',
+        action='store_true',
+        help='Skip CI when updating individual MRs when using batches'
+    )
     config = parser.parse_args(args)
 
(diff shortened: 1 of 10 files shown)


marge/batch_job.py
```diff
@@ -203,6 +220,11 @@ def execute(self):
         self._repo.checkout_branch(target_branch, 'origin/%s' % target_branch)
         self._repo.checkout_branch(BatchMergeJob.BATCH_BRANCH_NAME, 'origin/%s' % target_branch)
 
+        batch_mr = self.create_batch_mr(
```
bostik, line 223: This was somewhat surprising, because so far the batch MR creation has been guarded by `only_allow_merge_if_pipeline_succeeds` flag. Doing it here earlier works as well: if run without your new config options it's going to be used in the older sequence as expected.

bostik (commented):
You could fold this commit to its parent. It only adds noise to the commit history.

marge/app.py
```diff
@@ -213,6 +213,17 @@ def regexp(str_regex):
         action='store_true',
         help='Disable fast forwarding when merging MR batches'
     )
+    parser.add_argument(
+        '--use-merge-commit-batches',
+        action='store_true',
+        help='Use merge commit when creating batches, so that the commits in the batch MR '
+             'will be the same with in individual MRs. Please add sudo scope to the access token.\n',
```
bostik, line 220: If the `sudo` scope is needed for the functionality, we should change the wording used here. "Please add" implies it's nice to have, but not necessary. Maybe something like `Requires sudo scope in the access token.` ?

marge/batch_job.py
tclh123: Yep

PR #270, opened 2020-08-28
Retry requesting status up to 5 consecutive times when getting 5xx er…

valentindavid:
…rors

This should help gitlab in periods where gitlab has network
issues. That happens once in a while, and during this time there are
few 5xx errors (502 and 504). When one happens on a MR status request,
then Marge just unassign the MR, or close and re-opens a new batch.
When 5xx errors happen every 10 minutes, nothing gets merged.

### DIFF
--- a/marge/job.py
+++ b/marge/job.py
@@ -4,6 +4,7 @@ import logging as log
 import time
 from collections import namedtuple
 from datetime import datetime, timedelta
+import requests
 
 from . import git, gitlab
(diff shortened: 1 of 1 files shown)


qqshfox:
Thanks for fighting against the bad network. The retry facility would be even more useful if it could be extracted to a decorator.

nanonyme:
@qqshfox is that a requirement for this to be merged?

PR #271, opened 2020-08-31
Use target branch name in batch merge to allow multiple instances

### LINKED ISSUE #267: Cannot run multiple batch-enabled instances on separate target branches
In a project we use a marge bot instance per release branch so that we can run merges in parallel. However in that configuration, only one instance  can do batches. This is because the branch name used for the batch is constant: `marge_bot_batch_merge_job`.

Instead it should be using a name that depends on the target branch. So for example for target branch `release/20.08`, it should use `marge_bot_batch_merge_job/release/20.08`.

However there is no way to configure the name of the batch branch.

valentindavid:
A project can use multiple instances of marge bots assigned to
different target branches. This allows dealing with different
release branches in parallel.

However batches could not be used on multiple instances because
they conflicted on the name of the branch created for the batch.
By naming this branch after the target branch, then multiple
batch branches can exist.

Fixes #267


sg70:
@tclh123 Is there a chance that this PR gets merged?

tclh123:
@sg70 Yeah, I think this is good. Not sure if we want it to be configurable?

qqshfox:
@valentindavid Thanks for the contribution. Changing the branch name for branch will be a breaking change. Would you please add a configuration that is disabled by default?

PR #276, opened 2020-10-31
Fixed a bug which prevented marge batch merging with fast forward commits

felix-ulonska:
fix(BatchMergeRequest): fixed a bug which prevented marge batch merging with fast forward commits

batch_mr_sha was not set when fast forward merging


tclh123:
@felix-ulonska Thanks!

Issue #288, opened 2021-02-13
Is Nix worth the complexity?


brettdh:
I've been working on several enhancements currently deployed internally at my company, and I have more planned, all of which I'm hoping to submit as pull requests - but I have to say, the Nix-based build process is making it really hard to develop. 

Some examples:
- I'm currently trying to add a single package via pip, and running `pypi2nix` (which is now [abandonware](https://github.com/nix-community/pypi2nix)) took **24 minutes** to run. (In contrast, something like `poetry add` takes seconds.) Plus, it modified some files in ways that I don't understand, and I don't yet know if it worked.
- To build the docker image on MacOS, I have to run the nix build _inside a docker container_, because nix busybox package isn't supported on MacOS.
- I'm trying to tweak the docker configuration, and I know my way around a Dockerfile fairly well, but nix adds a new layer of complexity that I have to learn.

It seems like running in Docker is the most common use case. Is running in Nix a use case that still needs to be supported? If not, shedding that complexity would greatly simplify development for potential contributors like me.

My current plan (at some point) is to add a new plain-Dockerfile build process off to the side of the nix build. It's not a super complex app, so there's not a lot to duplicate. Hopefully I can make this a pull request to make this proposal more concrete.

nejch:
nix looks promising to me in general especially for OS dependencies, but I agree in this context it was easier for me to just test my stuff in a plain python container with pip-installed marge-bot and its deps for #289 (see this quick hack https://gitlab.com/nejch1/test-marge-bot-as-ci-service/-/blob/master/.gitlab-ci.yml#L7-8).

I also think poetry's lockfiles or something similar would probably get the job done now, I'm guessing nix was originally added to get around plain pip's poor management of transitive dependencies.

snim2:
I must admit that whenever I've raised a PR for this project I've just turned TravisCI on for my fork, rather than trying to navigate Nix, but perhaps that's just me!

qqshfox:
Hi @brettdh @snim2 , thanks for giving the detailed feedbacks. While I was trying Marge on my macOS, I found:

1. `nix-shell` doesn't even work on macOS. https://github.com/NixOS/nixpkgs/issues/91748 and https://github.com/NixOS/nixpkgs/issues/104580. Tried update the source to 20.09 but it didn't work either.
[...]

PR #289, opened 2021-02-14
Allow running marge-bot in CI pipelines or as a single CLI job

nejch:
This PR allows running marge-bot as a single job rather than looping indefinitely, so it can be managed via CI pipeline schedules instead of deploying it to your infrastructure.

See also renovate-bot's approach for something similar. https://docs.renovatebot.com/self-hosting/#gitlab-cicd-pipeline

I've added this as a simple flag since I wanted to keep codebase change to a minimum, although I was initially considering another `console_scripts` entrypoint. But I didn't want to change your approach to deployment too much (see also https://github.com/smarkets/marge-bot/issues/288) so I hope this is fine.

I also considered auto-detecting/defaulting to the `GITLAB_CI` and `CI_SERVER_URL` (for `gitlab_url`) env vars if marge detects it's in a CI environment, to get a bit closer to zero-config. But I also didn't want to break the existing response to required args. It's tempting though, let me know if you'd accept that :)

For an example run, see https://gitlab.com/nejch1/test-marge-bot-as-ci-service/-/jobs/1029017409.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,9 @@ optional arguments:
                            [env var: MARGE_SOURCE_BRANCH_REGEXP] (default: .*)
   --debug               Debug logging (includes all HTTP requests etc).
                            [env var: MARGE_DEBUG] (default: False)
+  --cli                 Run marge-bot as a single CLI command, not as a long-running service.
+                        This may be used to run marge-bot in scheduled CI pipelines or cronjobs.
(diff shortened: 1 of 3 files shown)


qqshfox:
I agree that we shouldn't break existing deployment. Thank you @nejch !

PR #291, opened 2021-02-23
Remove nix

### LINKED ISSUE #288: Is Nix worth the complexity?

qqshfox:
Resolves #288


brettdh:
I'm not a maintainer, but I suspect this would be more likely to be accepted if it _adds_ a plain Dockerfile rather _without_ removing the Nix setup. This could include removing the dockerize stuff from the Nix setup, but not removing the Nix setup altogether. As was discussed on #288, there are still users who prefer to deploy marge-bot via Nix.

nejch (commented):
I'd also say points 2 and 5 could be split into its own PR that gets merged first, to make the purpose of this PR clearer (basically cherry-pick https://github.com/smarkets/marge-bot/pull/291/commits/fdd7721840bfad2384e1daf83cf288b38f48201f but without the dependency upgrades).

And just my opinion but at this point GitHub Actions are probably nicer for quick CI tests than Travis ;) At least in my experience as of late 2020 Travis quality and job queues have deteriorated. Maybe not for deployment yet, as you have secrets stored already from what I see.

qqshfox:
@nejch thanks for the feedback. I've submitted a new MR to upgrade the dependencies and fix lint warnings introduced by the latest linters. However other parts, e.g. testing with more python versions, plain-Dockerfile, are all heavily based on the removal of nix. I'm not sure if it's viable to rewrite them along with the nix stuff.

nejch:
> @nejch thanks for the feedback. I've submitted a new MR to upgrade the dependencies and fix lint warnings introduced by the latest linters. However other parts, e.g. testing with more python versions, plain-Dockerfile, are all heavily based on the removal of nix. I'm not sure if it's viable to rewrite them along with the nix stuff.

Yeah that's basically what I had in mind for splitting up the linting issues, thanks :)

PR #293, opened 2021-02-25
Allow merging when ready, regardless of predefined order

nirizr:
In this PR I propose two additional flags, mostly relevant to non-batch mode:
1. ci-timeout-skip - Will change the behavior of the `ci-timeout` flag. The default behavior is that if a CI is running more than allowed by the ` ci-timeout` flag, Marge will give up on merging that MR and will reassign it from itself. This flags allows that instead of unassigning Marge will just skip that MR and try again at a later time. This also prevents Marge from hanging on one MR for the CI's duration. 
2. `skip-pending` - This may not be the best name. Normally, Marge (in non-batch mode) will only attempt merging the first MR within a project (giving up as mentioned above if CI takes too long). This flag will attempt merging all relevant MRs, in the right order.

Our CI process takes an hour to run, and is under-resourced so some MR may take several hours to pass CI.
The combination of the above two flags allowed us to set a very short (1-5 minutes) `ci-timeout` and merge MRs as they become ready to be merged, instead of waiting a long while for a single MR within a project. I think this also allows Marge to better handle multiple projects in a parallel-like manner, but I'm not certain.

WDYT?

Disclaimers:
1. I didn't yet check this PR but I've used something very similar for over 6 months now, I first wanna make sure the idea is acceptable and then will do what's required for it to be accepted.
2. Until very recently we've used version 0.9.2, so this approach can conflict with features added after that version.
3. I'm aware batch mode will also handle some of the issues that caused us to implement these flags but we rather avoid the batch mode for the time being.


marge/bot.py
```diff
@@ -166,13 +166,22 @@ def _process_merge_requests(self, repo_manager, project, merge_requests):
...
+            log.info('Attempting to merge the oldest MR...')
+            merge_request = merge_requests[0]
+            merge_job = self._get_single_job(
+                project=project, merge_request=merge_request, repo=repo,
+                options=self._config.merge_opts,
+            )
+            merge_job.execute()
+        else:
+            for i, merge_request in enumerate(merge_requests):
+                log.info('Attempting to merge MR #%d / %d of assigned MRs...', i + 1, len(merge_requests))
```
okainov, line 179: Why not use `enumerate(merge_requests, start=1)` instead? :)

PR #279, opened 2020-12-07
Rename Master role as Maintainer

snim2:
At some point GitLab renamed its user roles, and `Master` became `Maintainer`. This commit brings the marge-bot up to date with this change.

PR #294, opened 2021-02-27
AccessLevel enum matches GitLab docs

snim2:
Valid access levels, according to:

https://docs.gitlab.com/ce/api/access_requests.html

are currently:

* No access (0)
* Minimal access (5) (Introduced in GitLab 13.5.)
* Guest (10)
* Reporter (20)
* Developer (30)
* Maintainer (40)
* Owner (50) - Only valid to set for groups

This commit ensures that the enum matches the docs.
This is a follow-up to PR #279 -- sorry I missed this one at the time!


qqshfox:
Good catch. Thank you @snim2.

PR #297, opened 2021-03-24
allow squash in single merge request mode

bigc2000:
allow squash In MR.I am heavily use marge-bot,Thanks a lot!

### DIFF
--- a/marge/merge_request.py
+++ b/marge/merge_request.py
@@ -191,13 +191,14 @@ class MergeRequest(gitlab.Resource):
 
         raise TimeoutError('Waiting for merge request to be rebased by GitLab')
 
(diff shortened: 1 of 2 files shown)


snim2:
@bigc2000 I think this is something that can be set manually in the MR from the overview page. I can see the sense of having `marge-bot` squash every MR, but it does mean that you lose the more fine-grained control that GitLab offers.

If you do want to make squashing compulsory in a project, the **Settings -> General -> Merge Requests** page has an option to **Require** squashing.

bigc2000:
Emm, After  "squash" check button checked, but the no squash in MR finally. And same in batch mode,

PR #283, opened 2021-01-12
implement HTTPS support for cloning (#225)

matthiasbalke:
I finally found some time to implement this. I took @flaviouk idea and added a compatible argument parsing around it.
This is my first python project so I would love to get honest feedback, about parts that can be improved.

How do you like the idea of the mutally exclusive argument group of https vs. ssh?

Changes
* added configuration for HTTP(S) support
* added / configured cacerts nix package to allow SSL validation
* updated usage in Readme with latest `--help` output
* addded help section in Readme for configuring HTTP(S)

PR #308, opened 2021-05-11
Make CI work with GitHub Actions

nejch:
This will need two secrets configured by an admin in order for the DockerHub deploy to really work: `DOCKER_USERNAME` and `DOCKER_PASSWORD`, to replace encrypted secrets previously committed in the `.travis.yml` config.

This probably also needs an additional approval for every first contribution (see: https://github.blog/changelog/2021-04-22-github-actions-maintainers-must-approve-first-time-contributor-workflow-runs/) but here is an example run from my fork:
https://github.com/nejch/marge-bot/actions/runs/833323869

Helps with https://github.com/smarkets/marge-bot/pull/283 not running CI


matthiasbalke:
+1 for migrating to GitHub actions, which is a more modern and flexible build system. Thanks @nejch for your work!

qqshfox:
+1 for GitHub actions.

Do I need to configure/approve something to let actions triggered for this MR? I didn't find any jobs triggered.

nejch:
Thanks for really the quick feedback guys!

@qqshfox yes, I believe you need to "Approve and run". Do you see the setting in this PR as shown in the article below?
https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

Otherwise, perhaps Actions are disabled but I doubt it. Here are docs for enabling them:
https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow

Btw, you will need to do the Approve and Run every time a new contributor opens a PR for the first time, which is slightly annoying, but the "Auto-merge" option makes that better as you can approve and run, then auto-merge will just wait for the pipeline to succeed like in GitLab.

qqshfox:
Hmmm... Interesting... I don't find "Approve and run" anywhere. I even checked settings and it says `Allow all actions`. Probably the same reason (Github issue?) why travis isn't triggered?

qqshfox:
Will actions be triggered when lacking secrets? I'm still asking my colleague to configure DockerHub credentials.

nejch:
@qqshfox hmm that is strange, what happens if you pull my branch and push it to the main repo and open a draft PR, just to see if the problem is with forks? Maybe it needs to be merged first, but I'm surprised it doesn't work 😁

Generally actions should still run yes, but fail because of the missing environment variables (on master only, this only applies to the dockerhub job).

qqshfox:
Opened a PR and an action kicked in immediately. At the same time this PR now has a check referring to that action.

qqshfox:
Everything LGTM. I'll merge when DockerHub credentials configured.

qqshfox:
Fixed a syntax error in deploy action. Now it built and pushed the latest tag.

https://github.com/smarkets/marge-bot/commit/b5274b878a2db57a1d6542c5a3640741ae5b684e

nejch:
> 
> 
> Fixed a syntax error in deploy action. Now it built and pushed the latest tag.
> 
> [b5274b8](https://github.com/smarkets/marge-bot/commit/b5274b878a2db57a1d6542c5a3640741ae5b684e)

Oops, sorry had no way of testing that one without creating dummy stuff. Thanks :)

Issue #302, opened 2021-04-22
Batching Merge Requests doesn't work for forking workflow


sg70:
Marge-bot assumes during batching merge requests that the source branch is the origin repo when fusing the MR.
The source branch is not the origin repo when using the forking workflow. The results in an error message "merge: origin/source-branch - not something we can merge"

marge-bot version: 0.9.5

bigc2000:
So to me with bot 0.9.4,  and I modify  some code here, it maybe works
index a4b3479..d022ab3 100644
--- a/marge/batch_job.py
+++ b/marge/batch_job.py
@@ -110,17 +110,20 @@ class BatchMergeJob(MergeJob):
             if getattr(changed_mr, attr) != getattr(merge_request, attr):
                 raise CannotMerge(error_message.format(attr.replace('_', ' ')))
 
-    def merge_batch(self, target_branch, source_branch, no_ff=False):
+    def merge_batch(self, target_branch, source_branch, no_ff=False, source_repo_url=None, local=False):
         if no_ff:
             return self._repo.merge(
-                    target_branch,
-                    source_branch,
-                    '--no-ff',
+                target_branch,
+                source_branch,
+                '--no-ff',
+                source_repo_url,
+                local
             )
-
         return self._repo.fast_forward(
             target_branch,
             source_branch,
+            source_repo_url,
+            local
         )
 
     def update_merge_request(
@@ -177,6 +180,8 @@ class BatchMergeJob(MergeJob):
             merge_request.target_branch,
             merge_request.source_branch,
             self._options.use_no_ff_batches,
+            source_repo_url,
+            merge_request.source_project_id == merge_request.source_project_id
         )
         # Don't force push in case the remote has changed.
         self._repo.push(merge_request.target_branch, force=False)

PR #322, opened 2021-10-01
fix: resolve issue with batching MRs and forked MRs (#302)

sg70:
Since all forked MRs are prepared to local branches, the MR
acceptance procedure needs to be also local.

* fix batching MRs when source branch has different repo url
* add debug messages
* update unit tests
* add unit test for use case forked MRs
* remove credentials from debug log

Changes:
```
/tmpiq7raaob/tmpb84bjxf0 # git merge origin/dummy21 --ff --ff-only
merge: origin/dummy21 - not something we can merge
```

Into:
```
/tmpiq7raaob/tmpb84bjxf0 # git merge dummy21 --ff --ff-only
Updating b51f541..9525b89
Fast-forward
 21 | 0
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 21
```

Signed-off-by: Sascha Binckly <sascha.binckly@mbition.io>


sg70:
@qqshfox we have used this custom version for a couple of month without any issue. It would be nice to have this merged into the project.

@bigc2000 we created the issue and worked on a solution as well. We have our fix running internally since May. Let's team up on one PR.

marge/batch_job.py
```diff
@@ -225,6 +237,11 @@ def execute(self):
         )
         batch_mr_sha = batch_mr.sha
 
+        # delay in case Gitlab is slow to respond
+        waiting_time_in_secs = 20
+        log.debug('Waiting for %s secs for MR !%s to get status update', waiting_time_in_secs, batch_mr.iid)
+        time.sleep(waiting_time_in_secs)
```
qqshfox, line 243: Waiting for 20 seconds to get status update seems brutal. Would it be better to have a function to poll the status and have an interval and a limit? It can be used for other sleeps as well.

PR #323, opened 2021-10-01
feat: enable multiple target branches for batching MRs (#267)

sg70:
In order to support batch jobs in parallel, the batch branches
need to be tied to the target branch name.

Signed-off-by: Sascha Binckly <sascha.binckly@mbition.io>


sg70:
@qqshfox we have used this custom version for a couple of month without any issue. It would be nice to have this merged into the project.

marge/batch_job.py
```diff
@@ -20,19 +20,21 @@ class BatchMergeJob(MergeJob):
     def __init__(self, *, api, user, project, repo, options, merge_requests):
         super().__init__(api=api, user=user, project=project, repo=repo, options=options)
         self._merge_requests = merge_requests
+        # In order to run unit tests as standalone, initialize _batch_branch_name
+        self._batch_branch_name = BatchMergeJob.BATCH_BRANCH_NAME
```
qqshfox, line 24: Would you elaborate why this is needed please? Thanks.

sg70: The unit test do expect the fix name of the batch branch. The change does introduce a dynamic name of the batch branch depending on the target branch. Line 24 was a shortcut to have green tests.

Anyhow, I reworked the class. Class methods which do use the name of the batch branch did get an additional parameter (batch_branch_name). IMO this does make the code more readable. 
At the same time the initialization of class attribute is not nessary  anymore.

@qqshfox would you approve this PR now?

PR #325, opened 2021-10-12
Use default branch instead of master

### LINKED ISSUE #324: Assumes default branch is 'master'
When rebase fails, marge-bot attempts to checkout the default branch, which it assumes is named 'master':
https://github.com/smarkets/marge-bot/blob/master/marge/job.py#L348

This fails on repositories without a 'master' branch leading to uncaught exception. I think there should be a config setting for default branch name.

snim2:
Closes #324

### DIFF
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -10,6 +10,7 @@ INFO = {
     'path_with_namespace': 'cool/project',
     'ssh_url_to_repo': 'ssh://blah.com/cool/project.git',
     'merge_requests_enabled': True,
+    'default_branch': 'master',
     'only_allow_merge_if_pipeline_succeeds': True,
     'only_allow_merge_if_all_discussions_are_resolved': False,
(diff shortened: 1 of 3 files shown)


tests/test_project.py
```diff
@@ -10,6 +10,7 @@
     'path_with_namespace': 'cool/project',
     'ssh_url_to_repo': 'ssh://blah.com/cool/project.git',
     'merge_requests_enabled': True,
+    'default_branch': 'master',
```
mujx, line 13: Can't we use here the env variable `CI_DEFAULT_BRANCH` which is already provided by gitlab?

snim2: It's provided both by the env var and by [the API](https://docs.gitlab.com/ee/api/projects.html#get-single-project) so potentially we could do either, but the existing code seems to prefer the API over env vars, so I stuck with that

PR #335, opened 2022-01-23
Conditionally suffix batch brach based on parameter

nanonyme:
This attempts to retain backwards-compatibility and use batch branch suffixing when requested from command line


nanonyme:
Reasoning as https://github.com/smarkets/marge-bot/pull/271 for existing.

PR #336, opened 2022-01-26
Include more details when logging GitLabRebaseResultMismatch

perlun:
We are seeing this locally for whatever reason, _even though we are merging with margebot_ (e.g. multiple MRs in the same repo), and also in general whenever a branch is not up-to-date (=rebased on `master` in the main fork) when the MR gets assigned to the Marge-bot user. The aim here is to help to make debugging issues like this easier.

`marge-bot_1  | 2022-01-26 07:44:00,676 INFO Gitlab rebase didn't give expected result`


marge/single_merge_job.py
```diff
@@ -56,8 +56,8 @@ def update_merge_request_and_accept(self, approvals):
                     merge_request,
                     source_repo_url=source_repo_url,
                 )
-            except GitLabRebaseResultMismatch:
-                log.info("Gitlab rebase didn't give expected result")
+            except GitLabRebaseResultMismatch as err:
+                log.info("Gitlab rebase didn't give expected result: " + err)
```
perlun, line 60: ...or should this be `err.reason`? I don't work so much with Python so please help me get this right. The idea is to ensure that the exception message (as given in `synchronize_using_gitlab_rebase`) gets included in the log message here.

perlun:
Ping @snim2. Any chance someone could review this within the near future?

marge/single_merge_job.py
snim2: You can use `err.reason` here, yes. The definition of that class is:

```python
class CannotMerge(Exception):
    @property
    def reason(self):
        args = self.args
        if not args:
            return 'Unknown reason!'

        return args[0]

class GitLabRebaseResultMismatch(CannotMerge):
    def __init__(self, gitlab_sha, expected_sha):
        super().__init__(
            "GitLab rebase ended up with a different commit:"
            "I expected %s but they got %s" % (expected_sha, gitlab_sha)
        )
```

and the `@property` annotation means that you can use `reason` as an attribute.

snim2:
> Ping @snim2. Any chance someone could review this within the near future?

I've left you a comment, but I don't have rights to merge anything in this repo BTW!

marge/single_merge_job.py
perlun: Thanks, incorporated this suggestion now. :bow:

perlun:
Thanks @snim2, appreciated. :+1: Would @qqshfox be able to give this a look?

PR #338, opened 2022-04-07
Ignore jobs with manual trigger in CI

Knapock:
Marge bot gets stuck when there are manual jobs pending in CI.  
This PR assumes they are not necessary for a MR to be merged and considers the "manual" CI status as a valid status for merge.

### DIFF
--- a/marge/job.py
+++ b/marge/job.py
@@ -182,7 +182,7 @@ class MergeJob:
             if ci_status == 'skipped':
                 log.info('CI for MR !%s skipped', merge_request.iid)
                 return
-            
+
             if ci_status == 'manual':
                 log.info('CI for MR !%s has manual jobs', merge_request.iid)
                 return
--- a/marge/job.py
+++ b/marge/job.py
@@ -182,6 +182,10 @@ class MergeJob:
             if ci_status == 'skipped':
                 log.info('CI for MR !%s skipped', merge_request.iid)
                 return
+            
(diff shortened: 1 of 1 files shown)


alatiera:
>  This PR assumes they are not necessary for a MR to be merged and considers the "manual" CI status as a valid status for merge.

What we do instead, in gstreamer, in a couple of projects is to have manual jobs that are triggered by developers for testing, and a rule where we match on the mr assignee being marge and then make those jobs `when: always`. That way we use the manual status to avoid triggering expensive pipelines while the MR is still in its early stages and there might be a lot of iterations. So for our case manual is valid pipeline status.

Knapock:
> > What we do instead, in gstreamer, in a couple of projects is to have manual jobs that are triggered by developers for testing, and a rule where we match on the mr assignee being marge and then make those jobs `when: always`. That way we use the manual status to avoid triggering expensive pipelines while the MR is still in its early stages and there might be a lot of iterations. So for our case manual is valid pipeline status.

I had thought of that, but it can happen that no pipeline is triggered when a MR is assigned to Marge, because there is no rebase needed for example, and then this doesn't work. Our MRs get stuck until somebody else merges something and the MR needs to be rebased.

There's another PR in this repo that adds an option for Marge to trigger pending manual jobs. I guess I can build on that and add an option to ignore them, this could fit both our workflows.

PR #365, opened 2022-12-16
Support custom batch_branch_name from parameter

nanonyme:
If nothing is set, we use old default branch name. Possible solution for https://github.com/smarkets/marge-bot/issues/267


nanonyme:
@qqshfox are you still maintaining the project? This is more backwards-compat suggestion to same problem as https://github.com/smarkets/marge-bot/pull/271

marge/app.py
```diff
@@ -137,6 +137,12 @@ def regexp(str_regex):
         action='store_true',
         help='Enable processing MRs in batches\n',
     )
+    parser.add_argument(
+        '--batch-branch-name',
+        type=str,
+        default=None,
```
benjamb, line 143: You could have this default to `'marge_bot_batch_merge_job'`. It would simplify things in `BatchMergeJob.__init__()`.

However it would break the config check you've added below, but there were no previous checks for the use of batch specific configs without batching being enable.

nanonyme: Done.

PR #367, opened 2023-01-08
Update CA certificates to NSS 3.87

### LINKED ISSUE #356: LetsEncrypt cert from the included CA Bundle is expired
Following LetsEncrypt CA rotation in 2021, the ca-certs included in the latest image (0.10.1) are expired. I'm not familiar with Nix but it seems that `nixos-19.03` might be a very old base amd tje `cacert` package is outdated (`certifi` is also from 2019).

As a workaround, I've included my own CA in my deployment.

Is there a plan for https://github.com/smarkets/marge-bot/issues/288 and https://github.com/smarkets/marge-bot/pull/291 to be merged soon? Is this something I can contribute on?

If not, is there a plan to upgrade the version?

benkard:
Fixes #356.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,6 +1,25 @@
 let sources = import ./nix/sources.nix; in
-with import sources.nixpkgs {};
+with import sources.nixpkgs
 {
-  marge-bot = callPackage ./marge.nix {};
-  docker-image = callPackage ./dockerize.nix {};
+  overlays = [
+    (self: super: {
+
+      # Update NSS to a more recent version so we have an up-to-date
+      # CA certificate bundle.
+      nss =
+        self.callPackage
+          (import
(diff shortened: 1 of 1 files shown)


benkard:
This is something of a kluge, but it was the best I could come up with given my lack of knowledge about Niv. 

In my own testing I needed the following override in addition:

~~~diff
@@ -15,6 +15,9 @@ with import sources.nixpkgs
               sha256 = "sha256-aKGJRJbT0Vi6vHX4pd2j9Vt8FWBXOTbjsQGhD6SsFS0=";
             })
           { };
+
+      # Disable tests for Go because they fail for some reason.
+      go = super.go.overrideAttrs (superAttrs: { doCheck = false; });
     })
   ];
 };
~~~

I did not include it in the patch because it seemed out of place and I suspect an environment issue on my side.

PR #369, opened 2023-01-23
Fixes squash merge request for Single Job Mode

anuragagarwal561994:


anuragagarwal561994:
@snim2 can you help review these changes. Thanks

snim2:
> @snim2 can you help review these changes. Thanks

I'll have a look for you later today.

anuragagarwal561994:
@snim2 thanks, we have tested the main scenario in our test project and it is working. We are checking remaining scenarios.

Also I needed to add test case for the same, but would require some help to understand how it has been setup.

pylintrc
```diff
@@ -30,7 +30,7 @@ max-line-length=110
 [DESIGN]
 max-args=10
 max-attributes=15
-max-public-methods=35
+max-public-methods=40
```
snim2: This is possibly something that the project owners will want to look at, but IMO it's fine in this repo

snim2: Although this seems to be reverted in the next commit?

snim2:
I don't think I have maintainer rights here, but this looks OK to me!

pylintrc
anuragagarwal561994: Yes so there were 2 approaches in which this could have done. Earlier I was merging the project config with the MR config and then taking a decision.

But then I thought that there is only one case we need to address which is when at project level we have `sqaush_option: always`, so we are not taking any decision on behalf of the user or gitlab, instead we are respecting user's configuration and only sending `squash: true` when the project level configuration is set to always because in that case even gitlab won't allow the user to change the configuration at MR level.

For the earlier case I had to add a method to `merge_request` class which was exceeding the number of methods, but in this case we don't have to deal with this class at all, hence reverted.

marge/single_merge_job.py
```diff
@@ -92,9 +92,7 @@ def update_merge_request_and_accept(self, approvals):
 
             self.ensure_mergeable_mr(merge_request)
 
-            auto_squash = None
-            if self.is_auto_squash_enabled(merge_request):
-                auto_squash = True
+            auto_squash = True if target_project.squash_enforced else None
```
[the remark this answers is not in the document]
anuragagarwal561994: Again from my previous comment, a user is allowed to choose if he / she wants squash at 3 different places.
* in the project settings
* in the merge request settings
* while merging the merge request

Out of all the case only one case we had to solve for when squash is set to `required` at project level. In this case we are supposed to pass this extra parameter `squash: True` which more or less actually seems like a bug to me from `gitlab` side, remaining cases we are not sending this parameter and letting gitlab take the decision based on it's business logic based on user's input.

Hence a distinction between False & None. It has been sometime since I have worked in python, so not sure if this the best way to write it. But this is intentional, I can add a comment over this line to make it more clear.

theipster:
I've also tested this fix just now on one of our private GitLab repos (GitLab SaaS), and I can confirm this fix works quite nicely. ✅

anuragagarwal561994:
Thanks @theipster, @snim2 would require some help with the test cases, I haven't had experience writing test cases in python, I am trying still. There are few things I need to take care I believe:

* in `test_job.py`, I am mocking project which is making the condition

```
    auto_squash = (
            self._project.squash_enforced or
            merge_request.squash
        )
```

as always True, becuase squash_enforced is mocked. This is defeating the purpose for test case `test_ensure_mergeable_mr_squash_and_trailers`

Is there a better way this can be handled.


* I want to add test case for the same by changing the project level `always`, should I update the mock in this test case. No examples I guess to know how it is done.
* I wanted to add more cases in `test_merge_request.py` but I am getting confused with the flow, a little help I would be able to write the tests then.

theipster:
I'm neither an expert in Python nor marge-bot, but I can try and give some suggestions. 🙂 

_Firstly, this may help_: I consider the MR-level `squash` to mean "I **want to** squash", i.e. this is my choice, regardless of the consequences. This is in contrast to the project-level `squash_enforced` that means "I **need to** squash (otherwise the operation will fail)". Note that these are two different things, but are not mutually exclusive.

So, in terms of your questions:

> - in `test_job.py`, I am mocking project which is making the condition
>
>    ```
>    auto_squash = (
>            self._project.squash_enforced or
>            merge_request.squash
>        )
>    ```
>   as always True, becuase squash_enforced is mocked. This is defeating the purpose for test case `test_ensure_mergeable_mr_squash_and_trailers`

This is not defeating the purpose. With the introduction of this project-level setting, this simply means there are now _additional_ scenarios to consider:

| # | Scenario | Project `squash_enforced` | MR `squash` |
|---|-----------|---------------------|----------|
| 1 | "I **need to** squash" and so "I **want to** squash" | `True` | `True` |
| 2 | "I **need to** squash" but "I **don't want to** squash" (and suffer the consequence later) | `True` | `False` |
| 3 | "I **don't need to** squash" but "I **want to** squash" anyway | `False` | `True` |
| 4 | "I **don't need to** squash" and so "I **don't want to** squash" | `False` | `False` |

Scenarios 1 and 3 (MR `squash` is `True`) are already covered by the existing `test_ensure_mergeable_mr_squash_and_trailers` test case (maybe rename it to `test_ensure_mergeable_mr_squash_wanted_and_trailers` to make this clearer).

For scenario 2, you should add a new test case (e.g. `test_ensure_mergeable_mr_squash_needed_and_trailers`): configure the mock project `squash_option` to be `always`, configured the mock MR `squash` to be `False`, and then assert that `CannotMerge` is raised.

Scenario 4 is the "happy path" where we don't expect any exceptions. You can add a new test case for this if you like (this scenario is not specific to squashing).

> - I wanted to add more cases in `test_merge_request.py` but I am getting confused with the flow, a little help I would be able to write the tests then.

These existing `test_accept_*` tests do _behaviour verification_, whereby instead of asserting on the output (i.e. return values), it instead asserts on what actions (i.e. function calls) were triggered. This _behaviour verification_ pattern can be useful when it's painful to control the inner behaviour.

So, in our case, instead of asserting on the exact HTTP **response**, it simply asserts that a HTTP **request** was made with certain parameters. This is because it's painful to control the HTTP response.

In terms of how to what you need to do:

1. Firstly, fix the existing tests. The existing tests expect the HTTP request to be made with a `dict` of 3 entries, but you've now added an extra `squash = None` (default) to the `dict`, so the existing tests will fail. You'll need to update the assertions to reflect the new `dict`.
2. Now, add an extra test for each of the possible options that the `auto_squash` parameter can accept (I'm guessing `True`, `False` and `None`), and set the asserted `dict` respectively.

anuragagarwal561994:
@theipster thanks for this detailed explanation. I am currently busy with my work but would surely like to contribute and complete this PR by the end of this month with proper test cases.

theipster:
@anuragagarwal561994 I had some spare time and so I decided to roll up my sleeves and have a go myself - I hope you don't mind!

Please see https://github.com/anuragagarwal561994/marge-bot/pull/1, which adds some tests and also fixes some pip dependencies (which are unfortunately necessary for running the test suite).

anuragagarwal561994:
@theipster thanks a lot for doing this, I will review the changes from my end. No mind of course, I am not getting time to sit for this honestly :)

anuragagarwal561994:
@theipster I have found the changes to be appropriate and I have merged the same.

@snim2 I guess there is a dependent change by @theipster which is included in this MR as well
https://github.com/smarkets/marge-bot/pull/370

I suppose we can review and merge the same first and then we can re-review and merge this one.

anuragagarwal561994:
@snim2 did you get time to review and merge the same?

snim2:
@anuragagarwal561994 I don't have rights to merge in this repo :)

anuragagarwal561994:
@snim2 who can help with this?

snim2:
> @snim2 who can help with this?

Someone from Smarkets, I guess, but you'd have to look back through the old PRs and see who merged them to find a username

anuragagarwal561994:
@nithyashree675 @qqshfox can you help with the same?

PR #376, opened 2023-06-15
Adding Dockerfile

nithyashree675:
https://github.com/smarkets/marge-bot/issues/288


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.9
+
+COPY requirements_frozen.txt /requirements.txt
+RUN pip install -r requirements.txt
+
+COPY marge/ /marge/
+COPY marge.app /marge.app
+
+CMD ["/marge.app"]
```
dominik-horb-umg, line 9: Should this be `ENTRYPOINT` instead? We have a Helm chart to deploy marge bot and that doesn't work anymore with the `v0.11.0` Docker image because we just used to pass the missing args. Wouldn't be hard to fix for us, but at least looks like a breaking change.

